PR #5, opened 2017-01-25
Run UI tests on Firebase Test Lab

Sloy:

PR #6, opened 2017-01-26
Repeat the pager tests ten times in order to reproduce their flakiness

rocboronat:
Let's merge this PR and put it's code into #5 to avoid the flakiness of swipeBack. If the tests pass, then we could merge #5.


Sloy:
🖕

PR #14, opened 2017-02-14
Improve the Gradle compile line excluding Support Library

rocboronat:
As most projects use the Support Library, if we don't exclude it from the Barista dependency, the app version and the Barista one will conflict. Excluding it from Barista at the readme showcases this issue and the solution. If someone prefer to fix the issue by another way, they will be free to remove the exclude line and do their preferred magic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,9 @@ assertThatBackButtonClosesTheApp();
 # Download
 
 ```gradle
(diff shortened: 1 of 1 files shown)


Sloy:
Nice detail! This kind of conflicts are a big headache. 👍

rocboronat:
🎉

Issue #18, opened 2017-02-21
Add a way to do an elegant Thread.sleep


rocboronat:
https://youtu.be/isihPOY2vS4?t=11m25s

`uiController.loopMainThreadForAtLeast(millis)`

rocboronat:
Done at #19

PR #19, opened 2017-02-22
Add an elegant way to do a Thread.sleep()

rocboronat:
This is related with #18 

We learnt it here: https://youtu.be/isihPOY2vS4?t=11m25s


Sloy:
Great great great!

P.S.: Would have been nice if instead of just "don't use Thread.sleep(), use ours" he also explained why and what's the practical difference. But it seems like an improvement anyways.

👍

rocboronat:
Let's wait for Travis and merge it! :·)

PR #20, opened 2017-02-22
Add PermissionGranter to Barista

rocboronat:
As Barista is the last and better way to solve instrumental tests issues, we want to solve the issue related with allowing permissions with the system dialog. As Espresso can't tap that Dialog 'cause it's not an app one but a system one, we need to add UiAutomator.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -30,7 +30,6 @@ dependencies {
         exclude module: 'support-v4'
     }
     compile 'com.android.support.test.uiautomator:uiautomator-v18:2.1.2'
-    compile 'com.android.support:appcompat-v7:25.1.0'
 }
 
 publish {
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -29,6 +29,8 @@ dependencies {
     compile('com.android.support.test.espresso:espresso-contrib:2.0') {
         exclude module: 'support-v4'
     }
+    compile 'com.android.support.test.uiautomator:uiautomator-v18:2.1.2'
+    compile 'com.android.support:appcompat-v7:25.1.0'
 }
 
 publish {
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -30,6 +30,7 @@ dependencies {
         exclude module: 'support-v4'
     }
     compile 'com.android.support.test.uiautomator:uiautomator-v18:2.1.2'
+    compile 'com.android.support:support-annotations:25.1.0'
 }
 
 publish {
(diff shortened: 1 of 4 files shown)


library/build.gradle
```diff
@@ -29,6 +29,8 @@ dependencies {
     compile('com.android.support.test.espresso:espresso-contrib:2.0') {
         exclude module: 'support-v4'
     }
+    compile 'com.android.support.test.uiautomator:uiautomator-v18:2.1.2'
+    compile 'com.android.support:appcompat-v7:25.1.0'
```
Sloy: Is this dependency a must? It can cause dependency conflicts on users of the library. That's why we added the exclude above...

rocboronat: Yes, it's a must. It's related with this line of PermissionGranter:
`int permissionStatus = ContextCompat.checkSelfPermission(context, permissionNeeded);`

By the way... we should change the exclude that we recommend in the readme, 'cause if excluding this dependancy, PermissionGranter will fail...

Sloy: Look at the code of that method. It's one line with a nullcheck. Maybe we can copy that code and avoid the ContextCompat and remove the support library?

Sloy: Sorry, I could make your life easier. This is the code I mean:
```java
public static int checkSelfPermission(@NonNull Context context, @NonNull String permission) {
        if (permission == null) {
            throw new IllegalArgumentException("permission is null");
        }

        return context.checkPermission(permission, android.os.Process.myPid(), Process.myUid());
    }
```

rocboronat: Waaaw that's why I like you so much! Thanks-a-lot @Sloy :)

library/src/main/java/com/schibsted/spain/barista/permission/PermissionGranter.java
```diff
@@ -33,7 +34,7 @@ public static void allowPermissionsIfNeeded(String permissionNeeded) {
         }
       }
     } catch (UiObjectNotFoundException e) {
-      System.out.println("There is no permissions dialog to interact with");
+      Log.e("Barista", "setup");
```
Sloy: "setup"? Is it a mistake?

rocboronat: !!!!!!!

rocboronat: Fix'd :)

Sloy:
👍 !

PR #23, opened 2017-02-24
Adds FlakyActivityTestRule with @Repeat and @AllowFlaky annotations

Sloy:
This subclass of the ActivityTestRule let's you add two different annotations
to your test methods to execute them several times and either allow some flakyness
or avoid it.

TODO:

- [x] Add javadoc to the rule and annotations.
- [x] Explain on the readme
- [x] Replace the old `TenRepetitionsActivityTestRule` with this new one.


rocboronat:
Awesome, @Sloy !!! :·)

Sloy:
Hey @rocboronat take another look! I changed the implementation.

Changes:
- Fixes a bug where the default flaky attempts are applied also when there is a @Repeat. Now it only does it when both annotations are missing.
- The default flaky attempts are optional, you must call `allowFlakyAttemptsByDefault(n)` on the rule. To avoid deciding too much for the user. In our "CommonActivityRule" we can enable it always and be happy.
- Added some unit test for the logic of which statement is created.

rocboronat:
Lovely! Tons of poetry everywhere :·)

rocboronat:
🙌

Sloy:
@rocboronat all tasks done! I think it's ready for production now :D

sample/src/main/java/com/schibsted/spain/barista/sample/FlakyActivity.java
```diff
@@ -0,0 +1,13 @@
+package com.schibsted.spain.barista.sample;
+
+import android.os.Bundle;
+import android.support.v7.app.AppCompatActivity;
+
+public class FlakyActivity extends AppCompatActivity {
```
rocboronat: @Sloy what do you think about calling this Activity HelloWorldActivity? Actually, this Activity is not flaky at all, so we could use it for other tests, to avoid creating Activities everytime we need to test another feature in Barista. In addition, we could use this HelloWorldActivity to test that the text appears (but that's for other PR)

Sloy: Sounds good

rocboronat:
My comment there is just an opinion, so feel free to move on! 👍  By the way, launch a release if you have the time :·) I will start using this code instead of tailored one in several apps at Schibsted :·)

Sloy:
Suggestion applied. I will merge & release as soon as Travis agrees.

rocboronat:
Great! Thanks! :·)

rocboronat:
:tada:

PR #24, opened 2017-02-24
Clear preferences test rule

Sloy:
This rule lets us clear all preferences from the app between each test method. This way they are independent of any side-effects happening in any test.

TODO: 
- [x] Add Javadoc to the rule
- [x] Explain in the readme

PR #29, opened 2017-03-14
Clear Database Test Rule

Sloy:
This Rule works similarly to #24, but for SQLite databases.

It fetches all databases in the app, obtains all tables from them, and performs a "delete from" [without where](https://www.youtube.com/watch?v=i_cVJgIz_Cs).

I decided to do this instead of a DROP because I think it might break the app, not sure though.

This should be tested against different scenarios and sdk versions. I've seen in [Stetho](https://github.com/facebook/stetho/blob/master/stetho/src/main/java/com/facebook/stetho/inspector/database/DefaultDatabaseConnectionProvider.java) that they have different ways of opening databases depending on the sdk version. Wonder why.

TODO:
- [x] Check on different sdk versions, and fix any issues
- [x] Document on the javadoc and readme
- [x] (Won't do) _Maybe_ join with ClearPreferencesRule into one single rule for clearing app data.

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/cleardata/ClearDatabaseRule.java
+++ b/library/src/main/java/com/schibsted/spain/barista/cleardata/ClearDatabaseRule.java
@@ -1,0 +1,33 @@
+package com.schibsted.spain.barista.cleardata;
+
+import android.content.Context;
+import android.support.test.InstrumentationRegistry;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+/**
+ * This rule clears all app's Databases before running each test
+ */
+public class ClearDatabaseRule implements TestRule {
+
+  @Override
+  public Statement apply(final Statement base, Description description) {
+    return new Statement() {
+      @Override
+      public void evaluate() throws Throwable {
+        clearData();
+        base.evaluate();
+        clearData();
+      }
+    };
+  }
+
+  private void clearData() {
+    DatabaseOperations.clearAllDatabases(getAppContext());
+  }
+
+  private Context getAppContext() {
+    return InstrumentationRegistry.getTargetContext();
+  }
+}
--- a/sample/src/main/java/com/schibsted/spain/barista/sample/DatabaseActivity.java
+++ b/sample/src/main/java/com/schibsted/spain/barista/sample/DatabaseActivity.java
@@ -1,0 +1,71 @@
+package com.schibsted.spain.barista.sample;
+
+import android.app.Activity;
+import android.content.ContentValues;
+import android.content.Context;
+import android.database.Cursor;
+import android.database.sqlite.SQLiteDatabase;
+import android.database.sqlite.SQLiteOpenHelper;
+import android.os.Bundle;
+import android.view.View;
+import android.widget.TextView;
+
+public class DatabaseActivity extends Activity {
+
+  private TextView currentValueText;
+  private OpenHelper openHelper;
+
+  @Override
+  protected void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    setContentView(R.layout.activity_database);
+
+    openHelper = new OpenHelper(this);
+    currentValueText = ((TextView) findViewById(R.id.database_current_value));
+    findViewById(R.id.database_increment_button).setOnClickListener(new View.OnClickListener() {
+      @Override
+      public void onClick(View v) {
+        incrementValue();
+        showCurrentValue();
+      }
+    });
+
+    showCurrentValue();
+  }
+
+  private void showCurrentValue() {
+    SQLiteDatabase readableDatabase = openHelper.getReadableDatabase();
+    Cursor cursor = readableDatabase.query("User", new String[] { "name" }, null, null, null, null, null);
+
+    int currentValue = cursor.getCount();
+    currentValueText.setText(String.valueOf(currentValue));
+
+    cursor.close();
+    readableDatabase.close();
+  }
+
+  private void incrementValue() {
+    SQLiteDatabase writableDatabase = openHelper.getWritableDatabase();
+    ContentValues contentValues = new ContentValues(1);
+    contentValues.put("name", "Mr T");
+    writableDatabase.insert("User", null, contentValues);
+    writableDatabase.close();
+  }
+
+  private static class OpenHelper extends SQLiteOpenHelper {
+
+    public OpenHelper(Context context) {
+      super(context, "mydatabase", null, 1);
+    }
+
+    @Override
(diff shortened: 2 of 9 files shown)


Sloy:
One Issue found: It doesn't work if the database hasn't been created prior to the test (which will be the usual case). We must only try to delete databases if they exist. 🙌 

```
java.lang.RuntimeException: Unable to start activity ComponentInfo{com.schibsted.spain.barista.sample/com.schibsted.spain.barista.sample.DatabaseActivity}: android.database.sqlite.SQLiteException: no such table: User (code 1): , while compiling: SELECT name FROM User
#################################################################
Error Code : 1 (SQLITE_ERROR)
Caused By : SQL(query) error or missing database.
(no such table: User (code 1): , while compiling: SELECT name FROM User)
```

rocboronat:
Great work! Let's see how it progress!

By the way @Sloy , what do you think about having one Rule for every kind of thing to be remoded, and another one that just joins all of them? I think that having different rules will make them simpler. So, if we want to ease using all of them at the same tame (that will be the most common way to work with Barista), we can just create a Rule to join them. And... well, we can talk about it later, let's focus on this particular Rule :·D

Sloy:
That was my original idea. But I wasn't too sure. I thought about having just one for everything for simplicity, but if you think having both can be useful, let's go with it. I don't have a strong opinion on neither.

rocboronat:
@Sloy yes... the Rule that clears the Database needs 80 lines. The SharedPrefs one, 80 lines more I guess. Less have them smaller as possible, and create a new Rule to join these Rules.

By the way, at some point of the future, we will need to remove all .jpg and .png pictures of the filesystem (or whatever other extension) so... let's keep them isolated.

Sloy:
One rule to rule them all!! 💍

rocboronat:
:joy:

Sloy:
🍏  The issue I commented about clearing databases that don't exist is now fixed.
I changed the way of finding databases and opening them. I had to take into account the shadow database files like those ended in "-journal".

library/src/main/java/com/schibsted/spain/barista/cleardata/ClearDatabaseRule.java
```diff
@@ -0,0 +1,112 @@
...
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+/**
+ * This rule clears all app's Databases before running each test
+ */
+public class ClearDatabaseRule implements TestRule {
+
+  private static final String[] UNINTERESTING_FILENAME_SUFFIXES = new String[] {
```
rocboronat: Niiice!

sample/src/main/java/com/schibsted/spain/barista/sample/DatabaseActivity.java
```diff
@@ -0,0 +1,71 @@
...
+
+  private void incrementValue() {
+    SQLiteDatabase writableDatabase = openHelper.getWritableDatabase();
+    ContentValues contentValues = new ContentValues(1);
+    contentValues.put("name", "Mr T");
+    writableDatabase.insert("User", null, contentValues);
+    writableDatabase.close();
+  }
+
+  private static class OpenHelper extends SQLiteOpenHelper {
```
rocboronat: We could call it `OpenDatabaseHelper`. When I seen the dependencies of this Activity, I went crazy: What's this class?

Several horses and monkey died during my madness.

sample/src/main/java/com/schibsted/spain/barista/sample/DatabaseActivity.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public OpenHelper(Context context) {
+      super(context, "mydatabase", null, 1);
```
rocboronat: What means that `1`?

library/src/main/java/com/schibsted/spain/barista/cleardata/ClearDatabaseRule.java
```diff
@@ -0,0 +1,112 @@
...
+      String databaseFilename = databaseFile.getPath();
+      String sansSuffix = removeSuffix(databaseFilename, UNINTERESTING_FILENAME_SUFFIXES);
+      if (sansSuffix.equals(databaseFilename) || !originalAsSet.contains(new File(sansSuffix))) {
+        tidiedList.add(databaseFile);
+      }
+    }
+    return tidiedList;
+  }
+
+  private static String removeSuffix(String str, String[] suffixesToRemove) {
```
rocboronat: What do you think about calling this method `filterUninterestingSuffixes`? And I suggest to avoid passing the list of suffixest to remove. It seems that they will be always the same:  `UNINTERESTING_FILENAME_SUFFIXES`.

library/src/main/java/com/schibsted/spain/barista/cleardata/ClearDatabaseRule.java
```diff
@@ -0,0 +1,112 @@
...
+        clearData();
+        base.evaluate();
+        clearData();
+      }
+    };
+  }
+
+  private void clearData() {
+    for (File dbFile : tidyDatabaseList(getDatabaseFiles())) {
+      SQLiteDatabase database = performOpen(dbFile, 0);
```
rocboronat: What means this `0`?

library/src/main/java/com/schibsted/spain/barista/cleardata/ClearDatabaseRule.java
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Attempt to smartly eliminate uninteresting shadow databases such as -journal and -uid.  Note
+   * that this only removes the database if it is true that it shadows another database lacking
+   * the uninteresting suffix.
+   *
+   * @param databaseFiles Raw list of database files.
+   * @return Tidied list with shadow databases removed.
+   */
+  private static List<File> tidyDatabaseList(List<File> databaseFiles) {
```
rocboronat: This method name sounds strange... Don't know. I don't have a suggestion, by the way :·)

rocboronat:
There are some comments but... great work!

Sloy:
Rebased an squashed, much of the code on the rule has been changed @rocboronat.
I will check out your comments on the rest of the code later :)

rocboronat:
Nice, @Sloy ! Check my comments and go on applying the suggestion or merging :·)

Sloy:
Travis likes it. I tried a local build with InfoJobs project and everything works as expected.
Any final words @rocboronat?

rocboronat:
Ave, @Sloy , morituri te salutant! Great work there :·) It's merge time! Feel free to release a version if you need it! :·)

PR #31, opened 2017-03-15
Add a recycler view test method scrolls to a position

sergiocastilloscmspain:
scroll to a position in a recyclerView

### DIFF
--- a/sample/src/androidTest/java/com/schibsted/spain/barista/sample/RecyclerViewTest.java
+++ b/sample/src/androidTest/java/com/schibsted/spain/barista/sample/RecyclerViewTest.java
@@ -8,6 +8,7 @@ import org.junit.runner.RunWith;
 
 import static com.schibsted.spain.barista.BaristaAssertions.assertDisplayed;
 import static com.schibsted.spain.barista.BaristaRecyclerViewActions.clickRecyclerViewItem;
+import static com.schibsted.spain.barista.BaristaRecyclerViewActions.scrollTo;
 
 @RunWith(AndroidJUnit4.class)
 public class RecyclerViewTest {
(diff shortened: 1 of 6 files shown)


sample/src/androidTest/java/com/schibsted/spain/barista/sample/RecyclerViewTest.java
```diff
@@ -26,4 +27,10 @@ public void checkClickRecyclerViewItem_byPosition_atThree() {
     clickRecyclerViewItem(R.id.recycler, 3);
     assertDisplayed("Raspberry");
   }
+
+  @Test
+  public void checkScrollToRecyclerViewItem_byPosition_atThree() {
+    scrollTo(R.id.recycler, 3);
```
rocboronat: Is this test really testing that the scrollTo works? I mean, as the Recycler has only three items, the assertion will pass also if we don't call `scrollTo()`, isn't it?

sergiocastilloscmspain: yes 😅 
If you want, I could add more elements to this recycle to test the functionality

rocboronat: Yes! Add tons of elements, and then, do three tests: one to the first one, another test to an item of the center of the scroll, and another one to the last item of the list. In that way, we will be sure that it will always work :·)

By the way, thanks for the effort!

rocboronat:
@sergiocastilloscmspain here's my proposal! Now I'm running the whole test suite at localhost :·)

rocboronat:
@sergiocastilloscmspain green! Feel free to merge if you want :·) And thanks for the work!

PR #33, opened 2017-03-21
Fix a ## tag on readme

rocboronat:
@Sloy don't know why, readme was showing `## Dealing with Flaky tests` instead of <b>Dealing with Flaky tests</b>. This PR just removes your ## and retype it with my brand new bluetooth mac keyboard. It seems that my keyboard is better than yours 🌟


Sloy:
Hum weird. Atom stuff. Thanks for noticing it! 👍

Issue #34, opened 2017-03-21
Run task :library:mavenAndroidJavadocs on PRs


Sloy:
To avoid failures while deploying because of JavaDoc, we should also run the task `:library:mavenAndroidJavadocs` on PRs. Otherwise it will only run when we try to make a release, and fail miserably.

PR #35, opened 2017-03-21
Run upload task on PRs also

Sloy:
Using dry run and fake auth, of course.

This solves the problems found at #34, where some tasks are only run when releasing and might fail after a new PR has been merged already.

The original idea was to run the task that failed in my case (`: library:mavenAndroidJavadocs`), but I think using the dryRun feature of the publishing plugin is better. This way we make sure all release tasks are executed except for the upload itself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_script:
   - adb shell settings put global animator_duration_scale 0 &
   - adb shell input keyevent 82 &
 script:
-  - ./gradlew check
+  - scripts/bintrayUpload.sh
 deploy:
   - provider: script
     script: scripts/bintrayUpload.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_script:
   - adb shell settings put global animator_duration_scale 0 &
   - adb shell input keyevent 82 &
 script:
-  - scripts/bintrayUpload.sh
+  - scripts/build.sh
 deploy:
   - provider: script
     script: scripts/bintrayUpload.sh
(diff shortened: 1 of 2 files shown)


rocboronat:
Hey Rafa! Just seen the new .sh script but... how's that related with the main build at Travis? I mean, why Travis will run it?

Sloy:
@rocboronat you're right, I tried a few thing locally and didn't put them together properly 😂 

Now it should be fine. I changed the task `check` for `build`, I didn't see a reason not to. That way we make sure everything is working, right?

rocboronat:
Wait... now, we're uploading to bintray at every build? That's what we want?

```groovy
script:
  - scripts/bintrayUpload.sh`
```

rocboronat:
I think that we should call `scripts/build.sh` instead of `scripts/bintrayUpload.sh` at `script`

Sloy:
Today's not my day. Check again.
No more github for me until tomorrow 💀

rocboronat:
More by the ways... why not just call 

`./gradlew build bintrayUpload -PbintrayUser=dryrun-user -PbintrayKey=dryrun-key -PdryRun=true` at `script` ? It will save the need of opening another file to understand what's done on script. I mean, it's just one line... no need to add complexity there.

Sloy:
However you like. I did it in another file for consistency with the other script.

rocboronat:
I guess that it's to follow the style of `bintrayUpload.sh` ...? The thing is that hiding the magic of uploading to bintray is something good, cos it sounds that it will be the same on every project that works with this plugin, but knowing what's done on the build phase will be different for different projects. I mean, knowing how's done the upload is not important, but knowing what's happening on the `script` phase is interesting to be catchy.

By the way, feel free, is just an opinion and the change is very easy to do :·D

rocboronat:
Great! Thanks @Sloy :·) Feel free to squash and merge it! :·)

Issue #37, opened 2017-03-29
assertDisplayed doesn't work when multiple views matches the text


Sloy:
We found a case that I think Barista should handle.
In an Activity we had multiple views with the same text "Banana", and we were doing `assertDisplayed("Banana")` to verify that the data had been loaded and displayed.

Espresso fails because multiple views match the condition `withText()`.

From the perspective of someone using Barista and reading the method `assertDisplayed`, I think it should work. What do you guys think???

rocboronat:
Totally agree 😄

SmasSive:
Sorry but I don't understand what you are trying to say @Sloy ...
You are doing `assertDisplayed("Banana")` right? but Espresso fails... And AFAIK Barista is a wrapper of Espresso so, if Espresso fails, Barista should fail too.

But for sure I misunderstand something 💃

rocboronat:
Nope. The thing is why it's failing, @SmasSive . In the context @Sloy given, the app is showing "Banana" several times. As a Barista user, if I want to check if the text "Banana" is there, if there are lots of "Banana", `assertDisplayed("Banana")` should give a green, not an Exception.

SmasSive:
Yeah! Then I agree with you both :dancing_men:

rocboronat:
Then... it's PR time! :tada: hehehe

Sloy:
Exactly, sorry for the bad explanation 😅 

"Banana" appears twice in that screen, let's say it's the title of the screen and also a TextView inside the layout.

With espresso, if you run 
`onView(withText(text)).check(matches(isDisplayed()));`
and it finds two views with the text "Banana", it fails.

My point is: from the perspective of a Barista user, `assertDisplayed()` means it's displayed any number of times, the name of the method doesn't imply the other restriction.

Better @SmasSive ?

Sloy:
LOL I see I was too slow writing my comment, you both already answered :partyparrot:

rocboronat:
Welcome to the party, @Sloy ! Take some ice with you 🍸 🍹

SmasSive:
Jajaja @sloy now your explanation is perfect! 
So let's code it out!

Sloy:
Ideas?
We solved it by wrapping the `withText("Banana")` matcher inside a `firstOf(withText("Banana"))` matcher.
But it wouldn't work if the first banana is hidden and the second one is displayed.

Sloy:
[...]

PR #38, opened 2017-03-29
Accept repeated views

Sloy:
This is a solution proposal for #37 

The trick is to combine `firstViewOf()` with `allOf()`, and include the verification as a condition.
Yeah, that crazy. But it should work.

### DIFF
--- a/sample/src/androidTest/java/com/schibsted/spain/barista/sample/AssertionsTest.java
+++ b/sample/src/androidTest/java/com/schibsted/spain/barista/sample/AssertionsTest.java
@@ -50,6 +50,11 @@ public class AssertionsTest {
   }
 
   @Test
+  public void checkVisible_withRepeatedViews() throws Exception {
+    assertDisplayed("Repeated");
+  }
+
+  @Test
   public void checkInvisibleViews() {
     assertNotDisplayed(R.id.invisible_view);
     assertNotDisplayed(R.id.gone_view);
--- a/sample/src/androidTest/java/com/schibsted/spain/barista/sample/AssertionsTest.java
+++ b/sample/src/androidTest/java/com/schibsted/spain/barista/sample/AssertionsTest.java
@@ -52,6 +52,7 @@ public class AssertionsTest {
   @Test
   public void checkVisible_withRepeatedViews() throws Exception {
     assertDisplayed("Repeated");
+    assertDisplayed(R.string.repeated);
   }
 
   @Test
--- a/sample/src/androidTest/java/com/schibsted/spain/barista/sample/AssertionsTest.java
+++ b/sample/src/androidTest/java/com/schibsted/spain/barista/sample/AssertionsTest.java
@@ -51,6 +51,8 @@ public class AssertionsTest {
 
   @Test
   public void checkVisible_withRepeatedViews() throws Exception {
+    assertNotDisplayed(R.id.repeated_view_1_gone);
+
     assertDisplayed("Repeated");
     assertDisplayed(R.string.repeated);
   }
--- a/sample/src/main/res/layout/activity_main.xml
+++ b/sample/src/main/res/layout/activity_main.xml
@@ -13,10 +13,17 @@
       android:id="@+id/repeated_view_1"
       android:layout_width="wrap_content"
       android:layout_height="wrap_content"
+      android:visibility="gone"
       android:text="Repeated"
       />
   <TextView
       android:id="@+id/repeated_view_2"
+      android:layout_width="wrap_content"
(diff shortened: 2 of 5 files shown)


Sloy:
What do you think @rocboronat @SmasSive?
Is it too much hacky?

I think the only cons are that the error message in case of exception would be different, and that the Barista code gets more complex.

If you guys like it maybe we should apply this technique to other assertions as well.

rocboronat:
Wep @Sloy! Why not just catch the "MultipleMatchesException" and return green?

Sloy:
There could be multiple matches for withText but be hidden

SmasSive:
I like your idea, could be a little tricky but I like more than Roc's because catching an exception to do something expected... Usually it's not a good idea because you know, an exception it's something unexpected.
I don't have experience with espresso but for me, if the text is not visible it's like it's not there so if the first is not visible and the second one is, test should go green.
Does it help you? I hope so!

rocboronat:
@SmasSive all Barista is a try/catch library XD Actually, it manages this kind of Espresso things so, if we need to do bad things... we do them :·D

But, in this case, @Sloy is totally right: My idea was really bad :·D @Sloy , great solution! I can't figure anything better :·)

sample/src/main/res/layout/activity_main.xml
```diff
@@ -9,6 +9,25 @@
       android:layout_width="wrap_content"
       android:layout_height="wrap_content"
       android:text="@string/hello_world"/>
+  <TextView
+      android:id="@+id/repeated_view_1"
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:visibility="gone"
```
rocboronat: What do you think about adding a test that check that this TextView is invisible and the next ones visible? Just to check that the context that we're testing is the good one. If not, someone could remove this TextView, break your feature, and the tests will give a green instead of a red.

rocboronat:
By the way, to save you a click to Travis:

`/home/travis/build/SchibstedSpain/Barista/library/src/main/java/com/schibsted/spain/barista/BaristaAssertions.java:39:	Useless parentheses.`

rocboronat:
Going to do the simplest job ever seen... :D

rocboronat:
Mmmmm @Sloy , what do you think about doing the same at `assertDisplayed(int id)`?

sample/src/main/res/layout/activity_main.xml
Sloy: I like the idea. But I think it should be part of the "Given" of the new test, instead of a new test. As you said, it's to verify the context of this test, not something independent. Agree?

Sloy:
@rocboronat yes, it's part of my first comment :P 

> If you guys like it maybe we should apply this technique to other assertions as well.

rocboronat:
Op, your comment was wider enough to understand that you're talking about lots of other assertions, not the other `assertDisplayed()`. Please, be concise... we're techs 😝

sample/src/main/res/layout/activity_main.xml
rocboronat: Totally... the thing is that I don't know if it's common to check the context in the "given" part instead of setting it. But... let's give it a try :·)

Sloy:
@rocboronat ok let's just go with displayed for now.

Sloy:
Rebased with master! Definitive review?

sample/src/main/res/layout/activity_main.xml
rocboronat: Wep @Sloy how's the approach to this suggestion? It's something that we can do? Or it's too hard for this PR?

Sloy: Oh you're right. I thought I already did it! Let me add it real quick.

rocboronat: Thanks! :·)

Sloy: @rocboronat how about now?

sample/src/androidTest/java/com/schibsted/spain/barista/sample/AssertionsTest.java
```diff
@@ -50,6 +50,14 @@ public void checkVisibleViews_breaksWhenNeeded() {
   }
 
   @Test
+  public void checkVisible_withRepeatedViews() throws Exception {
+    assertNotDisplayed(R.id.repeated_view_1_gone);
```
rocboronat: Great way to check it! And so easy, I guess! :·D

rocboronat:
@Sloy great! Feel free to merge it! :·)

PR #40, opened 2017-03-30
Click button inside a Recycler's adapter

rocboronat:
This PR has not been too easy :·)

Tapping the item inside a row of an `Adapter` was not too hard, thanks to this StackOverflow question http://stackoverflow.com/questions/28476507/using-espresso-to-click-view-inside-recyclerview-item and an answer by http://stackoverflow.com/users/1270534/blade (thanks, **Blade**!)

By the way, adding this behavior broke some existing code. Why? When adding new buttons to the previously existing row, the top left corner of one of the button was appearing in the center of the row. As the usual `ViewActions.click()` doesn't use `perfomClick()` but just taps in the center of the view using a touch, that was clicking one of the new buttons instead of the row itself. For that reason, we needed to create a new `ViewAction` called `PerformClickAction` that does the click via `performClick()`.

The same happens on the old `ListView` actions, so it has been applied there, too :·)

**So**! This PR adds the behavior of its title and also fixes a bug that was affecting to previous `BaristaRecyclerViewActions.performClick()` method.


Sloy:
Hey I haven't read the code yet, but just a note:
I remember trying to use the `performClick()` action at InfoJobs to solve some problems with clicks, and we discarded it because it broke other things. I don't remember exactly what.

Just saying.

rocboronat:
@Sloy totally! I think that I made the change there :·D The key point is that we replaced ALL `ViewAction.click()` by `OurAction.performClick()`. That was a big change. In this case, we're using `performClick()` just to tap the body of the row instead of the buttons that it contain. That's what the real user does. The user doesn't tap the center of the row cos the corner of the button was there...

library/src/main/java/com/schibsted/spain/barista/custom/ClickChildAction.java
```diff
@@ -0,0 +1,61 @@
...
+
+      @Override
+      public void perform(UiController uiController, View view) {
+        View child = view.findViewById(id);
+        child.performClick();
+      }
+    };
+  }
+
+  public static ViewAction clickChildWithText(final String text) {
```
Sloy: How about naming these two methods just `clickChild()`? That's what we did in Barista with `assertDisplayed()`

library/src/main/java/com/schibsted/spain/barista/custom/ClickChildAction.java
```diff
@@ -0,0 +1,61 @@
...
+import android.view.View;
+import android.widget.TextView;
+
+import com.schibsted.spain.barista.util.ViewTreeAnalyzer;
+
+import org.hamcrest.Matcher;
+
+public class ClickChildAction {
+
+  public static ViewAction clickChildWithId(final int id) {
```
Sloy: Suggestion: use `@IdRes` on the parameter.

Sloy: Suggestion 2: It might make sense to support both R.id and R.string, similary to `assertDisplayed()`

library/src/main/java/com/schibsted/spain/barista/util/ViewTreeAnalyzer.java
```diff
@@ -0,0 +1,24 @@
...
+import android.view.ViewGroup;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class ViewTreeAnalyzer {
+  public static List<View> getAllChildren(View parent) {
+    if (!(parent instanceof ViewGroup)) {
+      return Collections.singletonList(parent);
```
Sloy: This returns the view itself. Doesn't that goes against the method name? I think it should return an empty list.

rocboronat: Well, the thing is that the code is recursive. Actually, you will call this code to get all the childs of a View, but actually what it does is to return all non ViewGroup things below it. If you can suggest a better name for the method...

library/src/main/java/com/schibsted/spain/barista/custom/ClickChildAction.java
rocboronat: Well, this is an Action, not part of the Barista's API. I mean, this is pure Espresso. It will be called from the internal Barista's code, so I prefered to be more explainatory than "catchy".

library/src/main/java/com/schibsted/spain/barista/custom/ClickChildAction.java
rocboronat: About the first, well catched! Going to do it!
About the second, it's not needed at the moment. Nice to have buuuut probably we could do it in another PR...

library/src/main/java/com/schibsted/spain/barista/util/ViewTreeAnalyzer.java
Sloy: But when the View is a ViewGroup, you don't include it on the list, right?

library/src/main/java/com/schibsted/spain/barista/custom/ClickChildAction.java
Sloy: Okai 👍

library/src/main/java/com/schibsted/spain/barista/util/ViewTreeAnalyzer.java
rocboronat: Yes, that's right. The key point is that if we just return an empty list as you suggested, we will not get the child of a child :·(

rocboronat:
Just solved merge conflicts using the GitHub website :·| GitHub guys rock.

PR #42, opened 2017-03-30
Add a method to click a radio button by position

rocboronat:
Here it comes! @Sloy as I needed a place to add Matchers, I used your class `HelperMatchers` at #38 . So, you will just add a method to that class if this PR is merged faster than yours :·)

By the way, the matcher this PR adds could be very useful for other tasks at Barista.

### DIFF
--- a/sample/src/androidTest/java/com/schibsted/spain/barista/sample/RadioButtonsTest.java
+++ b/sample/src/androidTest/java/com/schibsted/spain/barista/sample/RadioButtonsTest.java
@@ -2,12 +2,14 @@ package com.schibsted.spain.barista.sample;
 
 import android.support.test.rule.ActivityTestRule;
 import android.support.test.runner.AndroidJUnit4;
+
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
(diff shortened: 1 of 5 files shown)


sample/src/androidTest/java/com/schibsted/spain/barista/sample/RadioButtonsTest.java
```diff
@@ -38,4 +40,16 @@ public void checkRadioButtonsByText_secondItem() {
     clickRadioButtonItem(R.id.radiogroup, "Bye bye");
     assertDisplayed("" + R.id.second_item);
   }
+
+  @Test
+  public void checkRadioButtonsByPosition_firstItem() {
+    clickRadioButtonPosition(R.id.radiogroup, 0);
+    assertDisplayed("" + R.id.first_item);
```
Sloy: Why this assertDisplayed and this concatenation?

rocboronat: It's the same way we're checking that the ID of that item has been clicked. If you have a better idea... we could implement it in another PR, and change other methods as well :·D

Sloy: Nah

Issue #47, opened 2017-05-03
RecyclerView click actions don't work on wrapped views


Sloy:
The method `BaristaRecyclerViewActions.performClick()` only works when the `itemView` of the ViewHolder has the clickListener for the row.
If the clickListener is setted on a children view, the click is not dispatched to it.

For example, when the ItemView contains the clickable row AND a separator/shadow below. Or when using custom Adapters that wrap your view inside another one (like https://github.com/emilsjolander/StickyListHeaders).

The problem seems to be with `clickUsingPerformClick()` method, which calls `performClick()` on the specific `ItemView`, and doesn't let the click pass through its children.

rocboronat:
Fixed with #48 by @Sloy ! ❤️

PR #48, opened 2017-05-03
Fixes click on RecyclerView when ItemView is wrapped

Sloy:
This PR fixes the issue in #47.

On PR #40 we introduced a change to avoid clicking wrong things in a RecyclerView row, and allow clicking child views inside a row.

Unfortunately, that last change also broke the click when the ClickListener is set in a child view and not directly in the ItemView, as described in #47.

In this PR we:
- Reproduce the issue with a broken test (`WrappedViewClickTest`).
- Fix some flaky tests on RecyclerViewTest that were already giving false positives.
- Fix the original issue by **only** using `performClick()` when the view is clickable, or else use Espresso's `click()` action to allow propagating the click as usual.

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/custom/PerformClickAction.java
+++ b/library/src/main/java/com/schibsted/spain/barista/custom/PerformClickAction.java
@@ -6,6 +6,7 @@ import android.view.View;
 
 import org.hamcrest.Matcher;
 
+import static android.support.test.espresso.action.ViewActions.click;
 import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
 
 public class PerformClickAction {
@@ -32,7 +33,11 @@ public class PerformClickAction {
 
       @Override
       public void perform(UiController uiController, View view) {
-        view.performClick();
+        if (view.isClickable()) {
+          view.performClick();
+        } else {
+          click().perform(uiController, view);
+        }
       }
     };
   }
--- a/library/src/main/java/com/schibsted/spain/barista/custom/PerformClickAction.java
+++ b/library/src/main/java/com/schibsted/spain/barista/custom/PerformClickAction.java
@@ -36,8 +36,12 @@ public class PerformClickAction {
         if (view.isClickable()) {
           view.performClick();
         } else {
-          click().perform(uiController, view);
+          propagateClickToChildren(uiController, view);
         }
+      }
+
+      private void propagateClickToChildren(UiController uiController, View view) {
+        click().perform(uiController, view);
       }
     };
   }
--- a/sample/src/androidTest/java/com/schibsted/spain/barista/sample/RecyclerViewTest.java
+++ b/sample/src/androidTest/java/com/schibsted/spain/barista/sample/RecyclerViewTest.java
@@ -21,32 +21,32 @@ public class RecyclerViewTest {
   @Test
   public void checkClickRecyclerViewItem_byPosition_atTwo() {
     clickRecyclerViewItem(R.id.recycler, 2);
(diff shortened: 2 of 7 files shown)


sample/src/androidTest/java/com/schibsted/spain/barista/sample/RecyclerViewTest.java
```diff
@@ -21,32 +21,32 @@
   @Test
   public void checkClickRecyclerViewItem_byPosition_atTwo() {
     clickRecyclerViewItem(R.id.recycler, 2);
-    assertDisplayed("Avocado");
+    assertDisplayed("Avocado has been clicked");
```
rocboronat: Thanks :·)

library/src/main/java/com/schibsted/spain/barista/custom/PerformClickAction.java
```diff
@@ -32,7 +33,11 @@ public String getDescription() {
 
       @Override
       public void perform(UiController uiController, View view) {
-        view.performClick();
+        if (view.isClickable()) {
+          view.performClick();
+        } else {
+          click().perform(uiController, view);
+        }
```
rocboronat: As this is a bit of magic, we should explain why are we doing this trick. I don't have any suggestions but I feel that project newcomers will cry when they see this thing.

sample/src/androidTest/java/com/schibsted/spain/barista/sample/WrappedViewClickTest.java
```diff
@@ -0,0 +1,46 @@
...
+import static com.schibsted.spain.barista.custom.PerformClickAction.clickUsingPerformClick;
+
+public class WrappedViewClickTest {
+
+  @Rule
+  public ActivityTestRule<WrappedViewActivity> activityTestRule =
+      new ActivityTestRule<>(WrappedViewActivity.class);
+
+  @Test
+  public void click_on_button() throws Exception {
```
rocboronat: There are tons of tests written with the camel case style. If you prefer to start using other style, please, change all Barista's tests.

library/src/main/java/com/schibsted/spain/barista/custom/PerformClickAction.java
Sloy: Maybe if we extract the line ` click().perform(uiController, view);` to a method called `propagateClickToChildren()`?

rocboronat: Wow, **YES**! Great name!

sample/src/androidTest/java/com/schibsted/spain/barista/sample/WrappedViewClickTest.java
Sloy: Shit. Old habits.

library/src/main/java/com/schibsted/spain/barista/custom/PerformClickAction.java
Sloy: Done! 🎉

sample/src/androidTest/java/com/schibsted/spain/barista/sample/WrappedViewClickTest.java
Sloy: Done! 🎉

rocboronat: Party! 😄

rocboronat:
It's merge time! Feel free to bump a new version :tada:

Issue #49, opened 2017-05-04
Same Id matches multiple views in the hierarchy


despinola:
Having a view pager with multiple fragments, we are not able to click on one of the items because they have both the same id. 
When trying to get the item by id `ViewMatchers.withId(R.id.offer_list_recyclerview)` we get the following exception from espresso:
android.support.test.espresso.AmbiguousViewMatcherException:

The solution that we found is:
```
    sleep(1000);
    Espresso.onView(allOf(ViewMatchers.withId(R.id.offer_list_recyclerview), isDisplayed())).perform(new ViewAction[] {
        RecyclerViewActions.actionOnItemAtPosition(0, PerformClickAction
            .clickUsingPerformClick())
    });
```

With the isDisplayed() we filter the views to match only the ones that are on the screen, so other fragments doesnt influence. 
Since we are doing a swipe just before clicking, we need to wait until the swipe finish so there is only one visibile (thats why we put the sleep)

Enjoy.

rocboronat:
Thanks @despinola !

rocboronat:
Fix'd!

gilbertparreno:
It worked. Thanks!

rocboronat:
@gilbertparreno just to know, I understand you were having this using while using the last version of Barista?

PR #50, opened 2017-05-05
Reproduce issue #49 and fix it

rocboronat:
This PR just reproduces this issue https://github.com/SchibstedSpain/Barista/issues/49 and fixes it.

Thanks again for reporting it, @despinola !

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/BaristaRecyclerViewActions.java
+++ b/library/src/main/java/com/schibsted/spain/barista/BaristaRecyclerViewActions.java
@@ -7,10 +7,12 @@ import com.schibsted.spain.barista.exception.BaristaException;
 
 import static android.support.test.espresso.Espresso.onView;
 import static android.support.test.espresso.contrib.RecyclerViewActions.scrollToPosition;
+import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
 import static android.support.test.espresso.matcher.ViewMatchers.withId;
 import static com.schibsted.spain.barista.custom.ClickChildAction.clickChildWithId;
 import static com.schibsted.spain.barista.custom.ClickChildAction.clickChildWithText;
 import static com.schibsted.spain.barista.custom.PerformClickAction.clickUsingPerformClick;
+import static org.hamcrest.core.AllOf.allOf;
 
 public class BaristaRecyclerViewActions {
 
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/schibsted/spain/barista/BaristaRecyclerViewActions.java
```diff
@@ -24,21 +26,21 @@ public static void clickRecyclerViewItem(@IdRes int recyclerViewId, int... posit
...
 
   public static void clickRecyclerViewItemChild(@IdRes int recyclerViewId, int position, @IdRes int itemToClickId) {
-    onView(withId(recyclerViewId)).perform(
+    onView(allOf(withId(recyclerViewId), isDisplayed())).perform(
         RecyclerViewActions.actionOnItemAtPosition(position, clickChildWithId(itemToClickId)));
   }
 
   public static void clickRecyclerViewItemChild(@IdRes int recyclerViewId, int position, String text) {
-    onView(withId(recyclerViewId)).perform(
+    onView(allOf(withId(recyclerViewId), isDisplayed())).perform(
```
Sloy: Idea: We could have some static method that constructs this kind of matcher, to avoid duplication of code.
That way we only have the Barista magic in one place. (Maybe two places, because assertions need a different matcher I think).

But this PR can go to production as it is and make people happy already!

sample/src/androidTest/java/com/schibsted/spain/barista/sample/RecyclerViewInsideViewPagerTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static com.schibsted.spain.barista.BaristaAssertions.assertDisplayed;
+import static com.schibsted.spain.barista.BaristaRecyclerViewActions.clickRecyclerViewItem;
+import static com.schibsted.spain.barista.BaristaRecyclerViewActions.clickRecyclerViewItemChild;
+import static com.schibsted.spain.barista.BaristaRecyclerViewActions.scrollTo;
+
+@RunWith(AndroidJUnit4.class) public class RecyclerViewInsideViewPagerTest {
```
Sloy: Do we have a guideline for formatting of class annotations? I think this is the only place where the class name and the runner are in the same line.

rocboronat: I just installed AndroidSquare codestyle in my machine and it seems that it formats like this. But let me move to the previous version... Actually, it was hard to me to find the `public class` statement some minutes ago :·/

library/src/main/java/com/schibsted/spain/barista/BaristaRecyclerViewActions.java
rocboronat: Thanks @Sloy ! Just done it on #51 , that applies the same rule over all Barista Actions. By the way, right now, I don't think that this could be a good idea on Assertions, cos we're trying to be aggressive on the assertion part of the test. We can be happy on the when, but not on the end, if we want to make reliable tests.

But I can be wrong. Let's try that approach and fix it if it doesn't work! :tada:

rocboronat:
@Sloy done! Want to check it again?

library/src/main/java/com/schibsted/spain/barista/BaristaRecyclerViewActions.java
despinola: Sorry, I've not understood at all your explanation :(

rocboronat: @despinola Let me try to improve my English... hehehe @Sloy talked about creating a method that returns the `allOf(...)` Matcher, and talked about applying to Assertions.

Right now, we're only applying it to Actions, to fix the issue you described at #48 .

And then, focusing on applying the new Matcher on the Assertions, I think that it's not needed by now and, in the future, it could be an issue in the future. So I suggest to avoid applying the new suggested method to Assertions, cos it could produce worse tests.

rocboronat:
Thanks for asking @despinola ! I added explaining this magic at #7 , but, as #51 does the same thing than this PR but on all Barista's Actions, I propose to add it later :·)

library/src/main/java/com/schibsted/spain/barista/BaristaRecyclerViewActions.java
despinola: no no it's not your english.. I just needed more context :p 
now it's much better, thanks!

rocboronat: Thanks for asking, @despinola ! :·)

rocboronat:
💪

Issue #17, opened 2017-02-16
Change the ViewPager methods to apply only on displayed widgets


rocboronat:
In one case, we have a ViewPager inside a ViewPager. Then, the ViewPager that we want to move is repeated over lots of times inside another ViewPager. That makes Espresso to launch a "multiple ViewPagers match that ID". But well, obviously, we want to interact with the ViewPager that is visible.

So:

`onView(allOf(ViewMatchers.withId(R.id.the_id), ViewMatchers.isDisplayed())).perform(
            new ViewAction[] { ViewActions.swipeLeft() });`

rocboronat:
@Sloy I'll love to have feedback on this idea :·D

Sloy:
Wut? A ViewPager inside a ViewPager both with the same id? That sounds wrong :/

rocboronat:
Nope, a ViewPager that contains a page that has a ViewPager. So, the ID of the second ViewPager will appear in every page of the first ViewPager, so it will be repeated in the hierarchy tree.

rocboronat:
Something like this, where the car image is inside a small ViewPager that show lots of pictures of that car, while you can also change the page to take a look to another car:

![2017-02-16 17 13 34](https://cloud.githubusercontent.com/assets/456499/23032108/b3835438-f473-11e6-8842-e7da89f3dcc2.png)

If you want to swipe the images' ViewPager, you has match by ID and also look only for displayed ViewPagers.

Sloy:
Now I get it.
I don't have a strong opinion about this...
Seems like adding the isDisplayed is a hack for a specific case. And not only for the ViewPager methods. What about the buttons? Or any other view in that page?

rocboronat:
Yes, totally. It could apply to lots of things. But, excepting in the Buttons where we do the magic of scrolling until needed, when are you interested in tapping a button that is not visible? 

Something to thing about it. No hurries, by the way.

rocboronat:
Done at #51 !

PR #51, opened 2017-05-09
Just interact with visible things

rocboronat:
At #48, @despinola highlited an issue related with Recyclers inside ViewPagers. But #17 is related with the same issue. Actually, all BaristaActions* should only be applied to the visible item, to avoid issues with Fragments inside ViewPagers again and again.

This PR just adds the `isDisplayed()` matcher to all calls.

When this PR is applied, we will avoid having lots of `MultipleMatchingExceptions` when working with items inside ViewPagers.

* Actually, `scrollTo` needs to work on invisible widgets, too :·)

BEWARE THE ALIEN! This PR depends on #50

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/custom/DisplayedMatchers.java
+++ b/library/src/main/java/com/schibsted/spain/barista/custom/DisplayedMatchers.java
@@ -16,16 +16,16 @@ public class DisplayedMatchers {
 
   @NonNull
   public static Matcher<View> displayedWithId(@IdRes int id) {
-    return allOf(withId(id), isDisplayed());
+    return allOf(isDisplayed(), withId(id));
   }
 
   @NonNull
   public static Matcher<View> displayedWithText(@StringRes int text) {
-    return allOf(withText(text), isDisplayed());
+    return allOf(isDisplayed(), withText(text));
   }
 
   @NonNull
   public static Matcher<View> displayedWithText(String text) {
-    return allOf(withText(text), isDisplayed());
+    return allOf(isDisplayed(), withText(text));
   }
 }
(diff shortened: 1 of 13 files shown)


library/src/main/java/com/schibsted/spain/barista/custom/DisplayedMatchers.java
```diff
@@ -0,0 +1,31 @@
...
+import android.view.View;
+
+import org.hamcrest.Matcher;
+
+import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
+import static android.support.test.espresso.matcher.ViewMatchers.withId;
+import static android.support.test.espresso.matcher.ViewMatchers.withText;
+import static org.hamcrest.core.AllOf.allOf;
+
+public class DisplayedMatchers {
```
Sloy: Some javadocs to the class or methods might be helpful for users, to understand the magic of these matchers.

rocboronat: Wo... Actually, I think that the code is autoexplained. In addition, in this great example, the really nice method name also explains the implementation :joy:

`displayedWithId(int id) = allOf(withId(id), isDisplayed());`

rocboronat: Let me change the order of the matchers, to make them more readable if possible :·D

Sloy: Self-explaining code doesn't work for public APIs, IMHO ^^

rocboronat: But this code is not part of the public API, isn't it? It's code that we call from Barista. I know that someone can call it cos it's public, but... that's not what Barista aims to do.

In the other hand, the method does exactly what it explains. I mean, if I know how Espresso works, if I see a matcher called `displayedWithId(@ResId int id)`, it sounds that it matches with displayed views that had the given ID.

Let me skip adding the Javadoc. If it's a issue, we will add in as soon as it adds value to Barista users.

rocboronat:
Espresso tests passed! :tada:

PR #52, opened 2017-05-09
Introduce Barista's considerations that speed up the testing development

rocboronat:
Related with #7 
Waiting for #51

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,12 @@ public void some_important_test() throws Exception {
 }
 ```
 
+## Magic that Barista does for you
+
+In order to speed up testing, Barista keeps in mind some considerations.
+- **Scrolls when needed**: Filling a form inside a `ScrollView` requests you to scroll to the next item several times to avoid interacting with undisplayed widgets. To keep tests concise, Barista scrolls automatically before clicking a `Button` and before writing into an `EditText`.
+- **Just interacts with displayed Views**: Interacting with `Fragments` inside a `ViewPager` throws `AmbiguousViewMatcherException`, cos the widgets you interact with will be potentially repeated on different pages. Barista only interacts with displayed widgets, so you can focus on the behavior instead of wasting time on details.  
+
 # Download
 
 ```gradle
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ public void some_important_test() throws Exception {
 ## Magic that Barista does for you
 
 In order to speed up testing, Barista keeps in mind some considerations.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -154,6 +154,12 @@ public void some_important_test() throws Exception {
 }
 ```
 
+## Magic that Barista does for you
+
+In order to speed up testing, Barista keeps in mind some considerations.
+- **Scrolls when needed**: Filling a form inside a `ScrollView` requests you to scroll to the next item several times to avoid interacting with undisplayed widgets. To keep tests concise, Barista scrolls automatically before clicking a `Button` and before writing into an `EditText`.
```
Sloy: > before clicking a Button

Actually, clicking anything. This sentence might lead someone to think that it only works for buttons. Suggestion: Before clicking any View.

Also, I think it doesn't say clear enough how shitty is scrolling in Espresso.Here's an idea:
_"Interacting with Espresso in a `ScrollView` requires you to scroll to each view, which sometimes doesn't work the first time. Also trying to scroll outside a ScrollView makes produces an Exception, forcing you to change the test depending on the layout. To keep tests simpler. Barista scrolls....."_

It's probably too much text. But scrolling in Espresso is **so frustrating** that I think a small sentence doesn't give Barista enough credit.

README.md
```diff
@@ -154,6 +154,12 @@ public void some_important_test() throws Exception {
...
+- **Just interacts with displayed Views**: Interacting with `Fragments` inside a `ViewPager` throws `AmbiguousViewMatcherException`, cos the widgets you interact with will be potentially repeated on different pages. Barista only interacts with displayed widgets, so you can focus on the behavior instead of wasting time on details.
```
Sloy: Little detail: It's not just about Fragments. Any ViewPager with any kind of View inside that have the same layout have the same problem.

rocboronat: That's true!

rocboronat:
@Sloy please, feel free to improve this PR just committing into it. As your English is really better than mine, that will be easier for both 😄

README.md
rocboronat: Wep! Right now, the autoscroll is just done with Buttons and EditTexts :·/

rocboronat: Oh! Both were wrong. As you mentioned, all methods do the magic, in exception of `BaristaRadioButtonActions`. Opening an issue called #54!

rocboronat:
Wep! After having a lunch together... @Sloy 's suggestions applied verbatim!

PR #55, opened 2017-05-10
Autoscroll on RadioButtonsAction

### LINKED ISSUE #54: RadioButtons Actions doesn't scroll if needed

rocboronat:
As we explained at #52 , Barista is intended to scroll everytime is needed. But looking at tests and code, we noticed that it wasn't happening when clicking a RadioButton item.

This PR adds tests to check it and implements the solution of those tests.

Fixes #54

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/BaristaRadioButtonActions.java
+++ b/library/src/main/java/com/schibsted/spain/barista/BaristaRadioButtonActions.java
@@ -1,6 +1,8 @@
 package com.schibsted.spain.barista;
 
 import android.support.annotation.IdRes;
+import android.support.test.espresso.PerformException;
+import android.util.Log;
 
 import static android.support.test.espresso.Espresso.onView;
 import static android.support.test.espresso.action.ViewActions.click;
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/schibsted/spain/barista/BaristaRadioButtonActions.java
```diff
@@ -13,17 +14,20 @@
 public class BaristaRadioButtonActions {
 
   public static void clickRadioButtonItem(@IdRes int radioGroupId, @IdRes int itemToClickId) {
+    scrollTo(radioGroupId);
```
Sloy: In normal clicks we just do `.perform(scrollTo(), click())`. 
Are they different? Which one is better?

rocboronat: Actually, `Barista.scrollTo()` does the trick of doing the scroll 100 times to fix an unknown Espresso bug. And I guess that just perming an Espresso's scrollTo() doesn't work on some scenarios. But great question.

rocboronat:
Thanks for your feedback, @alorma !

PR #58, opened 2017-05-10
Create methods to avoid knowing the ID of the ViewPager if there's only one ViewPager

rocboronat:
The original idea? Look here! #8

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -31,6 +31,8 @@ dependencies {
     }
     compile 'com.android.support.test.uiautomator:uiautomator-v18:2.1.2'
     compile 'com.android.support:support-annotations:25.1.0'
+    compile 'com.android.support:support-core-ui:25.1.0'
+
     testCompile 'junit:junit:4.12'
     testCompile 'org.assertj:assertj-core:1.7.0'
 }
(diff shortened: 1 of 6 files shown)


library/src/main/java/com/schibsted/spain/barista/custom/DisplayedMatchers.java
```diff
@@ -28,4 +28,9 @@
   public static Matcher<View> displayedWithText(String text) {
     return allOf(isDisplayed(), withText(text));
   }
+
+  @NonNull
+  public static Matcher<View> displayedAssignableFrom(final Class<? extends View> clazz) {
+    return allOf(isDisplayed(), isAssignableFrom(clazz));
```
Sloy: Suggestion: A nice trick to avoid the ugly `clazz` hack is telling what it's for. Like `viewClass`.

library/build.gradle
```diff
@@ -31,6 +31,8 @@ dependencies {
     }
     compile 'com.android.support.test.uiautomator:uiautomator-v18:2.1.2'
     compile 'com.android.support:support-annotations:25.1.0'
+    compile 'com.android.support:support-core-ui:25.1.0'
```
Sloy: Q: Why is this new dependency for?

rocboronat: For ViewPager 😮

library/src/main/java/com/schibsted/spain/barista/custom/DisplayedMatchers.java
rocboronat: Nice! Actually, I copied it from the original `isAssignableFrom` method... but your idea is better :·)

library/build.gradle
Sloy: Aaaaaah I get it now.
Thanks!

Issue #62, opened 2017-05-15
Bug autoscrolling until next item at BaristaClickActions.click(id)


rocboronat:
This line introduces the issue: it doesn't scroll until the `View` is visible, cos just matches visible things.
`Espresso.onView(DisplayedMatchers.displayedWithId(id)).perform(new ViewAction[]{ViewActions.scrollTo(), ViewActions.click()});`

PR #64, opened 2017-05-16
Check that click() scrolls when needed

rocboronat:
As #62 explains, the autoscroll magic stopped working. We think that it happened at #51.

This PR reproduces the issue and fixes it.

But... it creates a new issue: it breaks the feature added at #51. So, we also test it and fix it.


sample/src/main/java/com/schibsted/spain/barista/sample/RadioButtonsActivity.java
```diff
@@ -25,7 +25,7 @@ protected void onCreate(Bundle savedInstanceState) {
     @Override
     public void onCheckedChanged(RadioGroup group, @IdRes int checkedId) {
       TextView textView = (TextView) findViewById(R.id.selected_item);
-      textView.setText("" + checkedId);
+      textView.setText(String.valueOf(checkedId));
```
rocboronat: This just fixes an AS warning

rocboronat:
Thank you @Sloy . I really appreciate that lovely words.

🏩

PR #73, opened 2017-05-19
Update logo

DiegoMillanR:
@rocboronat needed a new image for the library because he thought that they used was ugly. I was offered to participate in exchange for a bottle of Jack Daniel's. I hope to be paid


rocboronat:
Jack Daniels on the way! Thanks for the design, lovely designer!

PR #77, opened 2017-05-26
Try EditorConfig file to ease massive collaboration

rocboronat:
... but it's not as configurable as we need. If we want to use it, we need to change our checkstyle to match what EditorConfig generates. It just doesn't have as format customizations as we need.

I thought that the settings that appear in the main project readme were just a bit of all the settings it can manage... but nope. It's all it can do. Here's the list of powerups:
```
indent_style: “tab” or “space”
indent_size: a whole number
tab_width: a whole number
end_of_line: “lf”, “cr”, or “crlf”
charset: “latin1”, “utf-8”, “utf-8-bom”, “utf-16be” or “utf-16le”
trim_trailing_whitespace: “true” or “false”
insert_final_newline: “true” or “false”
```

As there's not a way to set the "continue line" spacing, one of the issues that we have is that this code:
```
UiObject allowPermissions = device.findObject(new UiSelector()
    .clickable(true)
```
is converted to:
```
UiObject allowPermissions = device.findObject(new UiSelector()
  .clickable(true)
```
And that makes checkstyle fail :·(

@Sloy @CristianGM any feedback there?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+root = true
+
+[*]
+
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false


Sloy:
That's a shame :( If it were just missing some settings, it would be better than nothing. But if it breaks existing rules as fundamental as your last example... I think it's a no-go 😞

rocboronat:
@Sloy what do you think about changing the checkstyle to accept that two spaces continue line instead of four? Actually, I personally accept both spacings. It's just a detail. Probably other rules will crash, but just for knowing, I'll investigate a little further.

Sloy:
I'm of the opinion that continuation lines should have different indentation than new lines, it makes the code less confusing. I think that's pretty standard.

From oracle documentation:

> Line wrapping for if statements should generally use the 8-space rule, since conventional (4 space) indentation makes seeing the body difficult. For example:
```
//DON'T USE THIS INDENTATION
if ((condition1 && condition2)
    || (condition3 && condition4)
    ||!(condition5 && condition6)) { //BAD WRAPS
    doSomethingAboutIt();            //MAKE THIS LINE EASY TO MISS
} 

//USE THIS INDENTATION INSTEAD
if ((condition1 && condition2)
        || (condition3 && condition4)
        ||!(condition5 && condition6)) {
    doSomethingAboutIt();
} 
```

rocboronat:
That's a really good example :·) Thanks for that!

Let me close the PR. I'll reopen it if it gets some awesommity in the future.

PR #82, opened 2017-05-30
Get the version from the tag at Git

alorma:
Use [grgit](https://github.com/ajoberstar/grgit) to automatic get version name & code from git, instead of manually updating README

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,10 +173,12 @@ In order to speed up testing, Barista keeps in mind some considerations.
 # Download
 
 ```gradle
-androidTestCompile('com.schibsted.spain:barista:1.1.0') {
+androidTestCompile('com.schibsted.spain:barista:x.y.z') {
   exclude group: 'com.android.support'
 }
 ```
+
+> `x.y.z` version can be found at the top of this README (badge)
 
 # License
 
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -1,5 +1,6 @@
 apply plugin: 'com.android.library'
 apply plugin: 'com.novoda.bintray-release'
+apply from: "$project.rootDir/tools/git-version.gradle"
 apply from: '../config/android-quality.gradle'
 
 android {
@@ -9,8 +10,8 @@ android {
     defaultConfig {
         minSdkVersion 9
         targetSdkVersion 25
-        versionCode 1
-        versionName "1.0"
+        versionCode gitVersionCode
+        versionName gitVersionName
     }
 }
 
@@ -44,7 +45,7 @@ publish {
     userOrg = 'schibstedspain'
     groupId = 'com.schibsted.spain'
     artifactId = 'barista'
-    publishVersion = '1.1.0'
+    publishVersion = gitVersionName
     desc = 'The guy who serves a great Espresso'
     website = 'https://github.com/SchibstedSpain/Barista'
 }
--- a/tools/git-version.gradle
+++ b/tools/git-version.gradle
@@ -1,0 +1,24 @@
+buildscript {
+    repositories {
+        jcenter()
+    }
+    dependencies {
+        classpath 'org.ajoberstar:grgit:1.9.2'
+    }
+}
+
+import org.ajoberstar.grgit.Grgit
+
+ext {
+    git = Grgit.open(currentDir: projectDir)
+    gitVersionName = ""
+    gitVersionName = git.describe()
+    gitVersionCode = git.tag.list().size()
+    gitVersionCodeTime = git.head().time
+}
+
+task printVersion() {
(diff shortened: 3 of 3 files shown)


tools/git-version.gradle
```diff
@@ -0,0 +1,24 @@
...
+        classpath 'org.ajoberstar:grgit:1.9.2'
+    }
+}
+
+import org.ajoberstar.grgit.Grgit
+
+ext {
+    git = Grgit.open(currentDir: projectDir)
+    gitVersionName = ""
+    gitVersionName = git.describe()
```
rocboronat: Why declaring gitVersionName twice?

README.md
```diff
@@ -173,11 +173,13 @@ In order to speed up testing, Barista keeps in mind some considerations.
 # Download
 
 ```gradle
-androidTestCompile('com.schibsted.spain:barista:1.1.0') {
+androidTestCompile('com.schibsted.spain:barista:x.y.z') {
```
Sloy: I personally hate with all my heart people that don't put the version in the copy&pastable dependency ^^

rocboronat: But you love people that automate things. There's a war inside your heart! ❤️

rocboronat:
@alorma are you going to be the first one to try this magic? 👍

library/build.gradle
```diff
@@ -9,8 +10,8 @@ android {
     defaultConfig {
         minSdkVersion 9
         targetSdkVersion 25
-        versionCode 1
-        versionName "1.0"
+        versionCode gitVersionCode
+        versionName gitVersionName
```
Sloy: You know... libraries don't really need versionCode and versionName fields because they're ignored, you can just remove them and be happy. That would also avoid the need of calculating a version code because it's not needed.

tools/git-version.gradle
Sloy: You can avoid adding another dependency by invoking git directly, and it's more similar to the actual git command. Like:
```
gitVersionName = 'git describe --abbrev=0 --tags HEAD^'.execute([], project.rootDir).text.trim()
```

library/build.gradle
rocboronat: Oh, true!

README.md
Sloy: This is the contrary of automating. This is telling the user: 
"Ha! You thought you could come here and copy&paste the dependency and be happy, right? Well, you don't! Go back to the top of the page, read the version name, remember it (because you can't copy it), paste the dependency in your project, come back here because you already forgot the version, and replace the x.y.z in your project with the actual version."

That's how I feel when I find something like this.

tools/git-version.gradle
Sloy: Ummm we're using the tag naming "v1.2.3". This commands returns it all, so using it as artifact version would make it `com.schibsted.spain:barista:v1.2.3` which is not what we want.
We either remove the prefix from `git describe`, or change the tag naming for the project.

README.md
rocboronat: Hahahaha that's true. Try to figure how to change the README.md too. We can do some sort of commits, replacing the old version with the new one. What do you think, @alorma ?

tools/git-version.gradle
rocboronat: Wow, thanks for catching it!

library/build.gradle
alorma: In fact is calculating it via the plugin... i can remove the lines, but...

tools/git-version.gradle
alorma: Why the `HEAD^` @Sloy ?

alorma:
The only way to update the README with the current tag name is, sadly, allowing travis make a commit, and that's not cool.

Sorry... i will rethink this

library/build.gradle
Sloy: What do you mean?

tools/git-version.gradle
Sloy: Ouch sorry, wrong copy&paste. `HEAD^` is to get the "previous" tag (it's what we use to obtain the changelog).

The actual command for the _current_ tag would be simpler: 
```
gitVersionName = 'git describe'.execute([], project.rootDir).text.trim()
```

alorma: In fact:

`git describe` will return a weird `v1.0.1-3-gf9b0032`

`git describe --tags` returns `v1.1.0`

So i think i will keep the `--tags` option

Sloy: describe only returns the weird hash when executing it in a commit that has no tag. Doing it in a tag returns just the tag name, which is fine.

PR #80, opened 2017-05-30
Run Espresso tests on Travis and adding a Rule to log to Travis

rocboronat:
This PR launches the emulator on Travis and runs the Espresso tests on it. It failed some months ago for some misconfiguration but, thanks to @CristianGM and @ferranpons and his work on https://github.com/SchibstedSpain/Leku, we can get that reliable configuration and apply it to Barista.

PR #78, opened 2017-05-26
Add menu click

alorma:
# Summary
Added menu click actions.

This pr tries to fix: https://github.com/SchibstedSpain/Barista/issues/72

# Changes

* Add `BaristaMenuClickActions`
   * Add actions to click by ID
   * Add actions to click by Text
* Add options inside click methods, for `Overflow items`

PR #83, opened 2017-05-30
Run Espresso tests on Travis

rocboronat:
Another way to implement #80 

In this case, we just add some noise to Travis log with the flags `--stacktrace --info`. This should make that Travis don't kill our process due to not writing into the log.

### DIFF
--- a/sample/src/androidTest/java/com/schibsted/spain/barista/sample/FlakyRuleTest.java
+++ b/sample/src/androidTest/java/com/schibsted/spain/barista/sample/FlakyRuleTest.java
@@ -51,7 +51,9 @@ public class FlakyRuleTest {
     }
   }
 
+  // WARNING: this test is flaky. Avoid running it at the C.I.
   @Test
+  @Ignore
   public void someDefaultFlakyTest() throws Exception {
     activityRuleWithDefaultFlaky.launchActivity(null);
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,14 @@ android:
   - android-25
   - extra
   - addon
+before_script:
+  - echo no | android create avd --force -n test -t android-21 --abi google_apis/armeabi-v7a
+  - emulator -memory 1024 -avd test -no-window -gpu off -no-boot-anim &
+  - android-wait-for-emulator
+  - adb shell input keyevent 82 &
 script:
   - ./gradlew build bintrayUpload -PbintrayUser=dryrun-user -PbintrayKey=dryrun-key -PdryRun=true
+  - ./gradlew :sample:connectedAndroidTest
 deploy:
   - provider: script
     script: scripts/bintrayUpload.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ android:
   - android-25
   - extra
   - addon
+  - sys-img-armeabi-v7a-google_apis-21
 before_script:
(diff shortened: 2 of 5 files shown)


rocboronat:
Hey @ferranpons ! It seems that this approach works as expected. By the way, the emulator is really, really slow passing tests... Let's see if the build ends :·)

rocboronat:
Methods added at #78 fail at the emulator :·( Let's try a simple `@AllowFlaky`...

rocboronat:
Well, as is, it works on my localhost, on emulators 19 and 21. Let's see what happens at Travis. If it crashes, I'll try with `@AllowFlaky` at `clickMenu()` tests.

rocboronat:
It works! :tada:

Sloy:
Why do we need to ignore menu tests?

rocboronat:
Let me forward you to the Travis log... but with some links to the exact lines ;·) Let fix that test on another PR.

checkMenuOnOverflowClick_byText() https://travis-ci.org/SchibstedSpain/Barista/builds/237567295#L7714

checkMenuOnOverflowClick_byId() https://travis-ci.org/SchibstedSpain/Barista/builds/237567295#L7718

checkMenuOnOverflowClick_byText() https://travis-ci.org/SchibstedSpain/Barista/builds/237567295#L7729

checkMenuOnOverflowClick_byId() https://travis-ci.org/SchibstedSpain/Barista/builds/237567295#L7733

rocboronat:
Some feedback about this, @Sloy ? And what do you think about the annotation's name?

sample/src/androidTest/java/com/schibsted/spain/barista/sample/FlakyRuleTest.java
```diff
@@ -51,7 +51,9 @@ public void someFlakyTest() throws Exception {
     }
   }
 
+  // WARNING: this test is flaky. Avoid running it at the C.I.
   @Test
+  @Ignore
```
Sloy: Don't you want to use your new annotation here? From the comment it looks like you do.

Sloy:
Well... the log is sooo long and big that the link to the exact line doesn't work 😹 
But other than that, I don't know :/ It's weird that only the menu tests are failing. I don't feel very confident.

The annotation name looks very explanatory, so that's OK.

Sloy:
Humm actually the tests that fail are the ones that open the overflow menu. Maybe it's something with the Android version?
Did you try running the tests locally in the same emulator version than Travis? Or with the same parameters? Maybe the `-no-window -gpu off -no-boot-anim` have something to do with this...

sample/src/androidTest/java/com/schibsted/spain/barista/sample/FlakyRuleTest.java
rocboronat: Nope. This test also crashes on localhost, so I prefer to keep it ignored on Travis AND in our computers. Newcomers doesn't know what's expected from this test, so it's strange to understand that a red is something randomly expected from this test. In addition, that's not related to passing the tests on Travis.

rocboronat:
@Sloy what do you think about fixing those tests in another PR? This PR runs 124 tests of 130 on Travis. It's the step forward that we expected since January 😅

Sloy:
@rocboronat sure, let's go ahead 👍

Issue #85, opened 2017-06-01
Create a rule to delete files


rocboronat:
Some apps (like Fotocasa one) save things on the disk. So, to isolate tests, we need to remove it before running the test.

Let's write a lovely Rule that removes the app files before running the test.

rocboronat:
Done!

PR #87, opened 2017-06-01
Add a Rule to remove files between running each test

rocboronat:
This PR will fix the issue #85 . @SmasSive will like it! :tada:


library/src/main/java/com/schibsted/spain/barista/cleardata/ClearFilesRule.java
```diff
@@ -0,0 +1,33 @@
...
+      public void evaluate() throws Throwable {
+        clearData();
+        base.evaluate();
+        clearData();
+      }
+    };
+  }
+
+  private void clearData() {
+    DatabaseOperations.clearAllDatabases(getAppContext());
```
alorma: I see a remove databases, but what about files created by app?

rocboronat:
Oh @SmasSive , it's a WIP! I'm currently working on it :·D

library/src/main/java/com/schibsted/spain/barista/cleardata/ClearFilesRule.java
rocboronat: Oh @alorma , it's a wip! I'm currently working on it :·D

rocboronat:
It's review time! In the meantime, I'm going to explain the feature at README.md

rocboronat:
... done! :·)

library/src/main/java/com/schibsted/spain/barista/cleardata/ClearFilesRule.java
```diff
@@ -0,0 +1,33 @@
...
+      public void evaluate() throws Throwable {
+        clearData();
+        base.evaluate();
+        clearData();
+      }
+    };
+  }
+
+  private void clearData() {
+    FileOperations.deleteAllFiles(getAppContext());
```
alorma: Uhm.

Context has access to many dirs:

captura de pantalla 2017-06-01 a las 16 55 51

What do you think allowing dir to be pass through constructor? So it will look like:

```java
File FILES_DIR = InstrumentationRegistry.getTargetContext().getFilesDir();
@Rule public ClearFilesRule clearFilesRule = new ClearFilesRule(FILES_DIR);
```

For example: i use `Context.getExternalCacheDir()` many times, and it could be great to delete files there too

library/src/main/java/com/schibsted/spain/barista/cleardata/FileOperations.java
```diff
@@ -0,0 +1,37 @@
...
+public class FileOperations {
+
+  static void deleteAllFiles(Context appContext) {
+    for (File file : getFiles(appContext)) {
+      file.delete();
+    }
+  }
+
+  private static Collection<File> getFiles(Context appContext) {
+    File filesDir = appContext.getFilesDir();
```
alorma: Same as before.

This file dir should be provided, otherway it can not use different dirs available on system

sample/src/main/AndroidManifest.xml
```diff
@@ -28,22 +28,25 @@
     <activity android:name=".DialogActivity"/>
     <activity android:name=".PreferencesActivity"/>
     <activity android:name=".DatabaseActivity"/>
+    <activity android:name=".FileActivity">
+      <intent-filter>
+        <action android:name="android.intent.action.MAIN"/>
```
alorma: Why need to change the main activity?

library/src/main/java/com/schibsted/spain/barista/cleardata/ClearFilesRule.java
rocboronat: What do you think if the rule just removes all the files of all the directories that don't need the API23 permission system?

library/src/main/java/com/schibsted/spain/barista/cleardata/FileOperations.java
rocboronat: Yes, let's keep that really important discussion in just one place, to keep brain sane :·D

sample/src/main/AndroidManifest.xml
rocboronat: At Barista, the main Activity is just a way to ease developers the task to develop an Activity that helps testing it. Please, don't focus the review on this change.

library/src/main/java/com/schibsted/spain/barista/cleardata/ClearFilesRule.java
alorma: Uhmmm

What about two options? One rule for all files, and one rule allowing to clear just a certain dir?

rocboronat: Just for being lean, I'm trying to do the smaller approach that fixes the issue we had yesterday at Fotocasa. If other apps need other approaches, we can do it on demand. But I feel that right now, a Rule that justs remove all the files is good enough...

alorma: :+1:

rocboronat: @SmasSive can you share where are you storing those magic files at the Fotocasa app? As there are lots of directories, I just need to be sure that the Rule will match your case, at minimum. Thanks!

rocboronat: @alorma are you still requesting changes? If not, please, remove your X :·P

rocboronat: Thanks @alorma !

rocboronat: Checked what Fotocasa does and, in the Use Case I'm trying to test, it saved things in `getCacheDir()`. So, this PR is useful enough :tada:

rocboronat: Oh, I'm working on a screen that saved things at the cache dir: `obj_is = new FileInputStream(new File(c.getCacheDir().getPath(), filename));`

Well, let's see how this Rule works in the real world. If we need to extend it, we will do it for sure :·)

Issue #71, opened 2017-05-18
Click by R.string.text


rocboronat:
Needed by @alorma and his great project at Schibsted :tada:

alorma:
Trying to fix in this PR: https://github.com/SchibstedSpain/Barista/issues/71

rocboronat:
Done by @alorma !

PR #88, opened 2017-06-01
Add feature: Click by string resource

alorma:
As issue:  [#71](https://github.com/SchibstedSpain/Barista/issues/71) points, we cannot click on a view wusing `R.sting.xxx` resources.

This PR fixes it!

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Barista introduces a discoverable API for the Espresso features. So, you and all
 ```java
 // Click widgets
 click(R.id.button);
+click(R.string.button_text);
 click("Next");
 clickBack();
 
--- a/library/src/main/java/com/schibsted/spain/barista/BaristaClickActions.java
+++ b/library/src/main/java/com/schibsted/spain/barista/BaristaClickActions.java
@@ -13,6 +13,7 @@ import static android.support.test.espresso.Espresso.pressBack;
 import static android.support.test.espresso.action.ViewActions.scrollTo;
 import static android.support.test.espresso.matcher.ViewMatchers.withId;
 import static android.support.test.espresso.matcher.ViewMatchers.withText;
+import static com.schibsted.spain.barista.custom.DisplayedMatchers.displayedAnd;
 
 public class BaristaClickActions {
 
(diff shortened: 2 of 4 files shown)


library/src/main/java/com/schibsted/spain/barista/BaristaClickActions.java
```diff
@@ -15,17 +18,43 @@
 
 public class BaristaClickActions {
 
-  public static void click(@IdRes int id) {
+  private static final ResourceTypeChecker RESOURCE_TYPE_CHECKER = new ResourceTypeChecker();
```
Sloy: It's probably a good idea to just make the methods static. Having an instance doesn't add any value, and we're doing this ugly thing in two places already.
Not sure if that belongs to this PR though.

library/src/main/java/com/schibsted/spain/barista/BaristaClickActions.java
```diff
@@ -15,17 +18,43 @@
...
       try {
-        scrollAndClickDisplayedView(id);
+        scrollAndClickDisplayedViewByString(id);
       } catch (PerformException parentIsNotAnScrollView) {
-        clickDisplayedView(id);
+        clickDisplayedViewByString(id);
       }
     } catch (PerformException parentIsNotAnScrollView) {
-      clickDisplayedView(id);
+      clickDisplayedViewByString(id);
```
Sloy: Hmm with this change each click method is not triplicated.
I think it's better to extract a method that receives a matcher, and pass the matcher from the generic method (`withText("")`, `withText(res)`, `withId(id)`).

Did I explained myself good enough? What do you think?

alorma: Uhmm, not understood you :'

alorma:
@rocboronat i will add on README.md as soon as i get it working in all situations, and fixed @Sloy comments

rocboronat:
Ok @alorma :·

alorma:
HI @Sloy i tried to implement your point, but then 2 tests fail (related with viewPager), could you hel with it?

Sloy:
@alorma yes, I think I know the issue. Let me try it and push some changes if it works.

alorma:
Zanku! 😘

Sloy:
There you go @alorma, with that small change tests are passing again.

I think the code can be simplified even more, but that's not related with the changes on this PR.

alorma:
Readme.md updated @Sloy 

Waiting for @rocboronat approve

README.md
```diff
@@ -16,6 +16,7 @@ Barista introduces a discoverable API for the Espresso features. So, you and all
 ```java
 // Click widgets
 click(R.id.button);
+click(R.stiring.button_text); // This will click a view that uses `<string name="button_text">Button text</string>`
```
rocboronat: There's a typo: `stiring`. In additon, we're introducing Barista to developers. Let's avoid overexplaining things like this... I mean, `click(R.string.something)` is going to click the view with the text `R.string.something` for sure.

alorma:
Fixed with amend commit! Thanks for typo!

Issue #26, opened 2017-03-02
Create one rule that uses the other Barista ones


rocboronat:
Just a way to ease copypasting to newcomers

rocboronat:
It could be great to also add the common `MockitoRule`

Sloy:
I'm on it!
We decided that it's better to not include MockitoRule because it's unrelated to espresso, the rule ordering doesn't matter, and it can cause issues with DexMaker (that's another story).

rocboronat:
Done at #90 by @Sloy !

PR #90, opened 2017-06-02
Create BaristaRule to rule them all

Sloy:
This rule includes all other rules, so it's easier and simpler to write tests without boilerplate.

It assumes sane defaults, and doesn't really allow customisation. Whoever wants to do something different, they can use the individual rules instead. Defaults are:
- Flaky attempts (without annotation): 10
- Launch activity automatically: false
- Initial touch mode enabled: true
- Clear preferences: always
- Clear databases: always

I wasn't sure how to add tests for it... Repeating the same tests that all rules have it's a lot of duplication. And I don't know how to test that a rule includes other rules.

I just tested manually by changing the current rules tests and replacing the original rules with this one.
It works.

This solves #26

TODO:
- [x] Explain in README

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
+++ b/library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
@@ -24,7 +24,7 @@ public class BaristaRule<T extends Activity> implements TestRule {
   private final ClearDatabaseRule clearDatabaseRule;
   private final FlakyActivityTestRule<T> activityTestRule;
 
-  public BaristaRule(Class<T> activityClass) {
+  private BaristaRule(Class<T> activityClass) {
     this.clearDatabaseRule = new ClearDatabaseRule();
     this.clearPreferencesRule = new ClearPreferencesRule();
     this.activityTestRule = new FlakyActivityTestRule<>(activityClass,
--- a/library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
+++ b/library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
@@ -3,6 +3,7 @@ package com.schibsted.spain.barista;
 import android.app.Activity;
 import android.content.Intent;
 import com.schibsted.spain.barista.cleardata.ClearDatabaseRule;
+import com.schibsted.spain.barista.cleardata.ClearFilesRule;
 import com.schibsted.spain.barista.cleardata.ClearPreferencesRule;
 import com.schibsted.spain.barista.flakyespresso.FlakyActivityTestRule;
 import org.junit.rules.RuleChain;
@@ -22,11 +23,13 @@ public class BaristaRule<T extends Activity> implements TestRule {
 
   private final ClearPreferencesRule clearPreferencesRule;
   private final ClearDatabaseRule clearDatabaseRule;
+  private final ClearFilesRule clearFilesRule;
   private final FlakyActivityTestRule<T> activityTestRule;
 
   private BaristaRule(Class<T> activityClass) {
     this.clearDatabaseRule = new ClearDatabaseRule();
     this.clearPreferencesRule = new ClearPreferencesRule();
+    this.clearFilesRule = new ClearFilesRule();
     this.activityTestRule = new FlakyActivityTestRule<>(activityClass,
         INITIAL_TOUCH_MODE_ENABLED,
         LAUNCH_ACTIVITY_AUTOMATICALLY)
@@ -38,6 +41,7 @@ public class BaristaRule<T extends Activity> implements TestRule {
     return RuleChain.outerRule(activityTestRule)
         .around(clearPreferencesRule)
         .around(clearDatabaseRule)
+        .around(clearFilesRule)
         .apply(base, description);
   }
 
--- a/library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
+++ b/library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
@@ -1,0 +1,55 @@
+package com.schibsted.spain.barista;
+
+import android.app.Activity;
+import android.content.Intent;
+import com.schibsted.spain.barista.cleardata.ClearDatabaseRule;
+import com.schibsted.spain.barista.cleardata.ClearPreferencesRule;
+import com.schibsted.spain.barista.flakyespresso.FlakyActivityTestRule;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+public class BaristaRule<T extends Activity> implements TestRule {
+
+  private static final int DEFAULT_FLAKY_ATTEMPTS = 10;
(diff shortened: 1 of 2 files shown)


Sloy:
Review time! I will update the readme in the meantime.

library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
```diff
@@ -0,0 +1,55 @@
...
+    this.clearPreferencesRule = new ClearPreferencesRule();
+    this.activityTestRule = new FlakyActivityTestRule<>(activityClass,
+        INITIAL_TOUCH_MODE_ENABLED,
+        LAUNCH_ACTIVITY_AUTOMATICALLY)
+        .allowFlakyAttemptsByDefault(DEFAULT_FLAKY_ATTEMPTS);
+  }
+
+  @Override
+  public Statement apply(Statement base, Description description) {
+    return RuleChain.outerRule(activityTestRule)
```
alorma: Is not better to add the `activityTestRule` at the end?

So all other rules are applied before activity is launched (clear DB, clear prefs, etc)

library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  private final ClearPreferencesRule clearPreferencesRule;
+  private final ClearDatabaseRule clearDatabaseRule;
+  private final FlakyActivityTestRule<T> activityTestRule;
+
+  public BaristaRule(Class<T> activityClass) {
+    this.clearDatabaseRule = new ClearDatabaseRule();
+    this.clearPreferencesRule = new ClearPreferencesRule();
+    this.activityTestRule = new FlakyActivityTestRule<>(activityClass,
```
alorma: Should be always Flaky test Rule? Not sure about this

library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
```diff
@@ -0,0 +1,55 @@
...
+import com.schibsted.spain.barista.cleardata.ClearPreferencesRule;
+import com.schibsted.spain.barista.flakyespresso.FlakyActivityTestRule;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+public class BaristaRule<T extends Activity> implements TestRule {
+
+  private static final int DEFAULT_FLAKY_ATTEMPTS = 10;
```
rocboronat: I suggest to move it to 5. It's a high value... and a 10 could slow failing tests too much. What's this value at InfoJobs?

Sloy: 7

library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
rocboronat: Why not?

library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
Sloy: The activity is NOT launched automatically, and that's an important point.
If you launch it automatically and use the ActivityRule as the inner rule, you can't use the FlakyActivityTestRule features (like repeating a test) because the ClearDataRule would not be applied between repetitions.

In general, it's better to use the ActivityRule as the innermost rule, and always launch the activity manually INSIDE the test method.

library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
Sloy: Well, that's the idea of the BaristaRule. To apply our common rules automatically without boilerplate.

library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
rocboronat: Let's stay on 10... All values are great. Increasing it adds reliability, and decreasing it gives fail-fast tests. But both are great. Let's move on with 10. We can change it if needed, so don't worry about this.

library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
alorma: I know i know!

What i mean is..  with this config, when you do `baristaRule.launchActivity()` it will launch:

* ActivityRule
* ClearDBRules
* ClearPrefsRule

is not better:

* ClearDBRules
* ClearPrefsRule
* ActivityRule

?

Sloy: That's not exactly true. The rules are not applied when you execute launchActivity, they're applied before.

```
-> Flaky activity rule
  -> ClearPrefsRule
    -> ClearDBRules
       test method{
         launchActivity()
       }
    <- ClearDBRules
  <- ClearPrefsRule
<- Flaky activity rule
```

The Flaky rule is the one that repeats tests, so it also repeats the inner rules.

alorma: great! thanks for the info

rocboronat:
@Sloy rememer to add the just created `ClearFilesRule` 😅

rocboronat:
And here's my another approval... :joy:

library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
SmasSive: Can we make it configurable?

library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
SmasSive: It's a great idea not launching the activity automatically, actually we don't use it so for us it's perfect! 💃

library/src/main/java/com/schibsted/spain/barista/BaristaRule.java
Sloy: Hmmm we can...
Do you really need/want to change it? I think it's a good default behaviour.
If you need it, I can add a secondary constructor.

SmasSive: No, I don't!
I just wanna put here some pace, candies and chocolate! ✌️ 🍬 🍫

Sloy: Let's go with the default behaviour then to keep things simpler.

Sloy:
Added to the readme, rebased with master and added the new ClearFilesRule.
@rocboronat @sergiocastilloscmspain any lasts words?

rocboronat:
I like trains 🚂 🚋 🚋 🚋  ... and also like this PR! Merge at will and launch a release. I'll use it on Fotocasa as soon as the release is published 💪

Sloy:
jcenter is down, apparently :(

SmasSive:
I like :potato: :potato: :potato:

PR #91, opened 2017-06-02
Stop running Espresso tests on Travis

rocboronat:
Passing tests on Travis emulator is not reliable as it should.

I made some tests running those tests on Firebase Test Lab, and they succeeds.

captura de pantalla 2017-06-02 11 03 31

So, this PR stops running the tests on Travis. Let's move #5 again!


Sloy:
We also had some flaky tests on the Firebase Test Lab, so let's be careful.

PR #93, opened 2017-06-06
Add Travis build stages

alorma:
Try to use: https://docs.travis-ci.com/user/build-stages

It keeps it more clean!

captura de pantalla 2017-06-06 a las 15 45 14

I've added 5 checks:

* Lint
* PMD
* Findbugs
* Checkstyle
* Bintray dryRun=true

and one Deploy stage

Someday, if we add UI test, we can add a new stage, or another check (to run it parallel).

Once gradle build cache is optimized to work across builds, we can add a stage at top to first build variants, and then apply checks


rocboronat:
@alorma just a feedback: it seems that builds are slower than before...

rocboronat:
Congrats, @alorma ! I think this will add value to the dev teams, but sadly, it builds in 77 minutes instead of just 12. I think that it's too time. Not just for the time waiting for the result of a PR, but also for wasting Travis unneeded resources at Travis. It makes the world a worse place due to the CO2 emissions. I think we should keep this PR in our mind, cos it's something great, but not for Barista right now. I know that you achieved an great achievement, but what do you think about this considerations?

alorma:
Totally agree.

The point here is that travis not runs all 5 steps in same moment. Maybe mix any of them?

we can keep this PR as #9C27B0 [blocked], and update it as gradle updates, maybe at some point it's reduced a lot

rocboronat:
Oh, feel free to change it, for sure! The think if that, right now, it doesn't add value as it could. For example, for me, it could be great to know if a code style issue is a checkstyle or a findbugs one without looking the Travis log. But if it has to get 15 more minutes to fail... actually, it's slowing me down :·/

So... yes, let's move it to blocked 🤝 Thanks for your contribution @alorma , I learned how new Travis Stages works by example 😅

rocboronat:
@alorma this PR is one month old. What do you think about closing it and leaving the branch alive?

PR #96, opened 2017-06-07
Add scrollTo() action that works with NestedScrollView

alorma:
This PR adds support on `scrollTo()` to `NestedScrollView`

I've also changed `click()` actions, so uses this new scroll, and we are available to click on NestedScrollView children

Thanks to @CristianGM for original Idea of create `BetterScrollToAction`

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/BaristaScrollToAction.java
+++ b/library/src/main/java/com/schibsted/spain/barista/BaristaScrollToAction.java
@@ -16,14 +16,14 @@ import static android.support.test.espresso.matcher.ViewMatchers.withEffectiveVi
 import static org.hamcrest.Matchers.allOf;
 import static org.hamcrest.Matchers.anyOf;
 
-public class BetterScrollToAction implements ViewAction {
+public class BaristaScrollToAction implements ViewAction {
   private final ScrollToAction scrollToAction;
 
-  public static BetterScrollToAction scrollTo() {
-    return new BetterScrollToAction();
+  public static BaristaScrollToAction scrollTo() {
+    return new BaristaScrollToAction();
   }
 
-  private BetterScrollToAction() {
+  private BaristaScrollToAction() {
     scrollToAction = new ScrollToAction();
   }
 
--- a/library/src/main/java/com/schibsted/spain/barista/BetterScrollToAction.java
+++ b/library/src/main/java/com/schibsted/spain/barista/BetterScrollToAction.java
@@ -1,0 +1,46 @@
+package com.schibsted.spain.barista;
+
+import android.support.test.espresso.UiController;
+import android.support.test.espresso.ViewAction;
+import android.support.test.espresso.action.ScrollToAction;
+import android.support.test.espresso.matcher.ViewMatchers;
+import android.support.v4.widget.NestedScrollView;
+import android.view.View;
+import android.widget.HorizontalScrollView;
+import android.widget.ScrollView;
+import org.hamcrest.Matcher;
+
+import static android.support.test.espresso.matcher.ViewMatchers.isAssignableFrom;
+import static android.support.test.espresso.matcher.ViewMatchers.isDescendantOfA;
+import static android.support.test.espresso.matcher.ViewMatchers.withEffectiveVisibility;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.anyOf;
+
+public class BetterScrollToAction implements ViewAction {
+  private final ScrollToAction scrollToAction;
+
+  public static BetterScrollToAction scrollTo() {
+    return new BetterScrollToAction();
+  }
+
+  private BetterScrollToAction() {
+    scrollToAction = new ScrollToAction();
+  }
+
+  @SuppressWarnings("unchecked")
+  @Override
+  public Matcher<View> getConstraints() {
+    return allOf(withEffectiveVisibility(ViewMatchers.Visibility.VISIBLE), isDescendantOfA(anyOf(
+        isAssignableFrom(ScrollView.class), isAssignableFrom(HorizontalScrollView.class), isAssignableFrom(NestedScrollView.class))));
+  }
+
+  @Override
+  public void perform(UiController uiController, View view) {
+    scrollToAction.perform(uiController, view);
+  }
+
+  @Override
+  public String getDescription() {
+    return scrollToAction.getDescription();
+  }
+}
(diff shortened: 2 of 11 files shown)


library/src/main/java/com/schibsted/spain/barista/BetterScrollToAction.java
```diff
@@ -0,0 +1,46 @@
...
+import android.widget.ScrollView;
+import org.hamcrest.Matcher;
+
+import static android.support.test.espresso.matcher.ViewMatchers.isAssignableFrom;
+import static android.support.test.espresso.matcher.ViewMatchers.isDescendantOfA;
+import static android.support.test.espresso.matcher.ViewMatchers.withEffectiveVisibility;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.anyOf;
+
+public class BetterScrollToAction implements ViewAction {
```
Sloy: "Better" is not a very objective and descriptive name 😂 
Don't we have a _better_ one?

library/src/main/java/com/schibsted/spain/barista/BetterScrollToAction.java
```diff
@@ -0,0 +1,46 @@
...
+
+  private BetterScrollToAction() {
+    scrollToAction = new ScrollToAction();
+  }
+
+  @SuppressWarnings("unchecked")
+  @Override
+  public Matcher<View> getConstraints() {
+    return allOf(withEffectiveVisibility(ViewMatchers.Visibility.VISIBLE), isDescendantOfA(anyOf(
+        isAssignableFrom(ScrollView.class), isAssignableFrom(HorizontalScrollView.class), isAssignableFrom(NestedScrollView.class))));
```
Sloy: From the original one I see they also include ListView. Probably we shouldn't remove it?

alorma: Uhmmm i don't see that:

```java 
  @SuppressWarnings("unchecked")
  @Override
  public Matcher<View> getConstraints() {
    return allOf(withEffectiveVisibility(Visibility.VISIBLE), isDescendantOfA(anyOf(
        isAssignableFrom(ScrollView.class), isAssignableFrom(HorizontalScrollView.class))));
  }
```

library/src/main/java/com/schibsted/spain/barista/BetterScrollToAction.java
alorma: Done

library/src/main/java/com/schibsted/spain/barista/BaristaScrollToAction.java
```diff
@@ -0,0 +1,46 @@
...
+import android.widget.ScrollView;
+import org.hamcrest.Matcher;
+
+import static android.support.test.espresso.matcher.ViewMatchers.isAssignableFrom;
+import static android.support.test.espresso.matcher.ViewMatchers.isDescendantOfA;
+import static android.support.test.espresso.matcher.ViewMatchers.withEffectiveVisibility;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.anyOf;
+
+public class BaristaScrollToAction implements ViewAction {
```
rocboronat: What do you think about explaining what this ViewAction exists? I mean, if it's to fix something related with the NestedScroll, we could clarify it in the name or something like that...

sample/src/androidTest/java/com/schibsted/spain/barista/sample/NestedScrollClickTest.java
```diff
@@ -0,0 +1,29 @@
...
+import org.junit.runner.RunWith;
+
+import static com.schibsted.spain.barista.BaristaAssertions.assertDisplayed;
+import static com.schibsted.spain.barista.BaristaClickActions.click;
+
+@RunWith(AndroidJUnit4.class)
+public class NestedScrollClickTest {
+
+  @Rule
+  public ActivityTestRule<NestedScrollViewActivity> activityRule = new ActivityTestRule<>(NestedScrollViewActivity.class);
```
rocboronat: Thanks for doing the test! ❤️

sample/src/main/res/layout/activity_nested_scroll.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
rocboronat: Why not a `FrameLayout`?

library/src/main/java/com/schibsted/spain/barista/BaristaScrollToAction.java
alorma: i don't think it's needed.

this action extends the Espresso one, and philosophy of Barista is to make tests easy for devs

if you see two `scrollTo` actions, you will choice Barista one, barista rules!

Also, i don't think is a good idea to explain in the name, because it may chance someday, if android (or support api) changes, should we change the name again?

`NestedScrollToAction` -> `NestedAndXxxxScrollToAction`?

sample/src/main/res/layout/activity_nested_scroll.xml
alorma: Why not `LinearLayout`? The test here is about `NestedScrollView` not the parent layout

rocboronat: Well, this LinearLayout has only one child. That's something that we need to avoid, isn't it?

library/src/main/java/com/schibsted/spain/barista/BetterScrollToAction.java
Sloy: Maybe you're seeing a different version than mine?
Here's the code in the repo: https://android.googlesource.com/platform/frameworks/testing/+/android-support-test/espresso/core/src/main/java/android/support/test/espresso/action/ScrollToAction.java#53

library/src/main/java/com/schibsted/spain/barista/BaristaScrollToAction.java
rocboronat: Yes. Actually, as it's an internal class, it's great to explain to other devs what it does. I mean, `BaristaScrollToAction` is the same than `AsdfghjkScrollToAction` 😂 As giving names is free, why not giving a good one? I suggest `NestedEnabledScrollToAction`. It's not the greatest name, but it expresses its intention.

As you note, if it changes in the future, let's say at 2042, the maintainers of the library will think in a better name for sure :joy:

library/src/main/java/com/schibsted/spain/barista/BetterScrollToAction.java
Sloy: Another approach is to re-use the `super.getConstraints()` to avoid re-implementing what espresso already does.

alorma:
Done @rocboronat

rocboronat:
Tests pass on my machine. Gogogo @alorma ! Feel free to also launch a release with this change :·)

Issue #74, opened 2017-05-21
Permission Error


kamontat:
```Java
@Test
public void validateInputText() {
	PermissionGranter.allowPermissionsIfNeeded(Manifest.permission.WRITE_EXTERNAL_STORAGE);
	onView(withId(R.id.fragment_input_id_number)).perform(typeText(SIMPLE_INT)).check(matches(withText(SIMPLE_INT)));
}
```

# description

I setting app to request permission first the first time running, I saw the dialog is out, But this code didn't click allow, and I cause `NoActivityResumedException` 

# Image

![screen shot 2560-05-21 at 2 25 38 pm](https://cloud.githubusercontent.com/assets/14089557/26281997/c9e55412-3e31-11e7-9abf-0888f50b02eb.png)

kamontat:
OK, I already fixed it. 

I change method `allowPermissionsIfNeeded` to...
```Java
public static void allowPermissionsIfNeeded(String permissionNeeded) {
	try {
		if (Build.VERSION.SDK_INT >= 23 && !hasNeededPermission(InstrumentationRegistry.getTargetContext(), permissionNeeded)) {
			UiDevice e = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation());
			UiObject allowPermissions = e.findObject((new UiSelector()).clickable(true).checkable(false).index(1));
			if (allowPermissions.exists()) {
				allowPermissions.click();
			}
			e.wakeUp();
		}
	} catch (UiObjectNotFoundException var3) {
		Log.e("Barista", "There is no permissions dialog to interact with", var3);
	} catch (RemoteException e) {
		Log.e("Barista", "Remote error", e);
	}
}
```
I add `e.wakeUp()` and remove `Sleep(long)`

rocboronat:
Hi @kamontat ! Can you send a PR, please? It will be great to know that you authored that change :·)

And one question, what device are you using to run the tests? Which API level?

Thanks a lot!

rocboronat:
Note for the future: probably related with #61

kamontat:
I test on pixel with Android 25 and also on Nexus5 Android 22 too

kamontat:
I have seen that error came because of sleep method

rocboronat:
Ok! And what about `e.wakeUp();`? It's really needed?

rocboronat:
Hey @kamontat , does Barista 1.5.1 fix this issue? If yes, please close the issue :·) Thanks!

[...]

PR #99, opened 2017-06-20
Fix permission granter

### LINKED ISSUE #61: PermissionGranter doesn't work on Android 23 or 25
Reproduced with QuitNow! tests on the emulator. PermissionGranter just avoids tapping the accept button.

Sloy:
We were using a sleep() which uses a ViewAction to make a magic efficient sleep. But since the permission dialog is on top of the application, Espresso cannot run the ViewAction.

Soooo changing it to a Thread.sleep() that doesn't require any view makes it work.

This fixes #61 and probably #74


rocboronat:
Nice! Thanks for this lovely test! :·)

PR #101, opened 2017-06-21
Remove api limit to PermissionGranter

Sloy:
The PermissionGranter class is annotated with RequiresApi 18 (because UI automator requires that api level).
This means that using PermissionGranter in an application that supports APIs lower than 18, Lint will give an error.

But in sdks lower than 23 we don't even use UiAutomator because permissions are granted at install time, so we're actually fine.

This PR removes the annotation so that we don't mislead the user.
I've tried running the PermissionGranter test on an emulator with api 17 and it works just fine.

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/permission/PermissionGranter.java
+++ b/library/src/main/java/com/schibsted/spain/barista/permission/PermissionGranter.java
@@ -4,7 +4,6 @@ import android.content.Context;
 import android.content.pm.PackageManager;
 import android.os.Build;
 import android.support.annotation.NonNull;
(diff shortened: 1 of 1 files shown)


rocboronat:
Nice! Thanks for the explanation, @Sloy 🎉

Sloy:
Jcenter is down, Travis is sad :(

rocboronat:
How are you? 😍

Sloy:
I'm high because of the smell of kerosene, so no worries :D

rocboronat:
Hum! Nice.

PR #102, opened 2017-06-23
Remove @IdRes from click() actions

alorma:
Since `BaristaClickActions.click()` supports both `R.id` and `R.string` we don't need anymore the `@IdRes` annotation


rocboronat:
Thanks for spotting! 👍

PR #103, opened 2017-07-06
Move one class and delete another one to keep the house clean

rocboronat:
This PR is just to keep the house clean. I found some issues when preparing some slides to talk about Barista... and here's the PR 🥇


rocboronat:
BTW, passed tests on local and they worked 💪

Sloy:
Question: Is this class a public API?

Because changing the package of a public API will break the client implementation.

rocboronat:
@Sloy nope, it's just an internal class. It's used by the public API `BaristaScrollActions.scrollTo()`.

Sloy:
Good!

rocboronat:
🎉

PR #109, opened 2017-07-18
Add stacktrace to log on failed flaky attempts

Sloy:
When using the @AllowFlaky rule or similar, Barista retries the test several times to see if it pass sometime.
At each attempt we print a log telling which attempt are we starting and the result of each one.

Since Espresso is so flaky and so many different things can go wrong, while debugging tests it can be helpful to
see what error caused each failed attempt.

This PR adds the stacktrace of the exception to the log.

This is how it looks like:
screen shot 2017-07-18 at 14 46 37


rocboronat:
Wow, nice @Sloy !

PR #114, opened 2017-07-27
Update to Espresso 3

Sloy:
I had to update the support library from 25.1.0 to 25.4.0, and change how we exclude some transitive dependencies.
I had to add the new Google Maven repository, where Espresso 3 is published.

I extracted all shared versions (libraries and sdks) to the main gradle file, because it was a nightmare trying to sync all versions from the library and sample.
This way future changes will be less painful.

With the migration to Espresso 3 there was just one code compilation error, because Espresso changed its included Guava package.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ android:
   - tools
   - tools
   - platform-tools
-  - build-tools-25.0.1
+  - build-tools-25.0.3
   - android-25
   - extra
   - addon
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,10 @@ android:
   - android-25
   - extra
   - addon
+
+  licenses:
+    - 'android-sdk-license-.+'
+
 script:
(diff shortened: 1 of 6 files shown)


Sloy:
Oops I have to confirm if the google maven repository is required by the consumers of barista or not. If it is, we need to update the readme.

alorma:
Does it means that all consumers will need to use espresso 3?

Sloy:
@alorma let me rephrase that: "all consumers will automatically start using espresso 3"

Sloy:
I can confirm: consumers need to include the maven repository in order to compile:

```
repositories {
    maven { url "https://maven.google.com" }
}
```

rocboronat:
Great work @Sloy ! If it passes all tests, let's move forward! About the maven thing, you could add it to the readme and also on the release itself, to keep the users noticed. Great great workkkkKKKK!

Issue #69, opened 2017-05-17
Migrate to Kotlin


Sloy:
Well it's obvious, isn't it?

There's too much Java in this project.

SmasSive:
Done!

PR #121, opened 2017-08-24
Add Kotlin to Barista! :tada:

SmasSive:
Welcome to the futu... Oh wait! To the presenter! 🍷

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,10 +1,12 @@
 buildscript {
-    repositories {
+  ext.kotlin_version = '1.1.4-2'
+  repositories {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.2'
+        classpath 'com.android.tools.build:gradle:2.3.3'
         classpath 'com.novoda:bintray-release:0.3.4'
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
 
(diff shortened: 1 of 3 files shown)


sample/build.gradle
```diff
@@ -25,8 +26,12 @@ dependencies {
...
 
+  compile "org.jetbrains.kotlin:kotlin-stdlib-jre7:$kotlin_version"
+
   testCompile 'junit:junit:4.12'
 
   androidTestCompile project(':library')
-
+}
+repositories {
+  mavenCentral()
```
Sloy: I don't think that adding mavenCentral() is needed for the subprojects. We already add jcenter() to all subprojects from the main build.gradle file.

build.gradle
```diff
@@ -1,10 +1,12 @@
 buildscript {
-    repositories {
+  ext.kotlin_version = '1.1.4-2'
+  repositories {
```
Sloy: Did the indentation of `repositories` change? Is it wrong now or was it wrong before?

SmasSive: AFAIK it uses 4 spaces indentation while library uses 4 and sample uses 2. The code is indented to 2.
I didn't want to change too much lines but I think is necessary a little reformat on those two gradle files. What do you think?

sample/build.gradle
SmasSive: Yes, you are right!

rocboronat:
BTW, this PR is related with #69

build.gradle
Sloy: Ok let's not worry about it here

PR #125, opened 2017-08-29
Migrate ClearDatabaseRule to Kotlin

Sloy:
I refactored the responsabilities of the rule and the DatabaseOperations file. The latter is now a simple collaborator to abstract from Android's SQLite database, instead of doing all the work.
The rule logic is now on the rule itself.

I added unit tests for the behavior of the rule. It wasn't really possible before because DatabaseOperations was made of static methods.

_"Why?"_ You may ask. 
Because I want to introduce a new feature in the rule, and I think this previous refactor will help a lot making a clean implementation.


library/src/main/java/com/schibsted/spain/barista/cleardata/ClearDatabaseRule.kt
```diff
@@ -0,0 +1,48 @@
...
+        @JvmField
+        internal val UNWANTED_FILENAME_SUFFIXES = arrayOf("-journal", "-shm", "-uid", "-wal")
+    }
+
+    override fun apply(base: Statement, description: Description): Statement {
+        return object : Statement() {
+            override fun evaluate() {
+                clearDatabases()
+                base.evaluate()
+                clearDatabases()
```
SmasSive: Why do you need to call twice this method? Looks weird...

Sloy: - Before: Clear the database so our test runs on a clean environment.
- After: Clear the database to leave the environment clean for the next test.

If everyone did it you could just do it at the beginning or at the end. But since tests are isolated and can't trust the other tests, it's good to make sure.

Sloy:
Travis fails, but hopefully #126 will go through soon.

Sloy:
Two grey checks count as one green? :D

PR #127, opened 2017-08-30
Allow excluding tables from the ClearDatabaseRule

Sloy:
This change adds a method to the rule that let's the user exclude tables matching a certain regular expression.
Useful for integration tests where some tables on the database are expensive to refill, and aren't modified between tests. Like static information.


Sloy:
This PR depends on #126

Sloy:
@SmasSive yes please!

PR #130, opened 2017-09-27
Fixes click with scroll and multiple views

Sloy:
This magic behavior of the click method was implemented, but didn't have any test covering it.
And guess what. It didn't work!

The `scrollAndClickDisplayedView` was using the `isDisplayed` condition to filter the button from the current viewpager. But since the button is below a scroll, isDisplayed would never match.

I created the missing test to reproduce it, and fixed the method by using a more complex view matcher. The `isDisplayed` filter is applied on the scrollable ancestor of the button, and not the button itself.

I had to change the try/catch logic to make it work as expected after the change.


Sloy:
Oops, spoilers!
screen shot 2017-09-28 at 09 36 58

Let me rebase and amend that.

rocboronat:
Nice, great work @Sloy !

PR #134, opened 2017-10-05
Improve error handling on magic methods

### LINKED ISSUE #116: Avoid calling the FailureHandler for expected errors
Barista does lots of magic tricks and we love that. But one of those tricks is to rely on `try/catch` patterns to allow higher level tests.

This has one big drawback: Espresso's [FailureHandler](https://developer.android.com/reference/android/support/test/espresso/FailureHandler.html) get's invoked every time, **even when the thrown exception is catched and ignored**.

Usually it's not a big deal, but when using custom FailureHandlers to do things like taking a screenshot when a test fails, it becomes a real pain. You can't tell when there was an actual error and when it's just a "ghost error" from Barista.

## Proposal
My proposal is to use the `withFailureHandler()` method on all Barista "magic" methods to override the global FailureHandler and avoid invoking it.

```java
try {
      onView(withId(id))
          .withFailureHandler(baristaFailureHandler)
          .check(matches(isDisplayed()));
    } catch (AssertionFailedError | PerformException e) {
      ...
    }
```

## Discussion
But there's something I'm not sure about. From the test perspective, what should happen when a 3-attempts test fails? The global FailureHandler should be notified somehow (to take the screenshot, for example). But it takes two parameters, a `Throwable` and a `Matcher<View>`. Since Barista usually tries with multiple matchers, which one should we use? Any of them? The simplest one? A fake one? 

I would love to hear some opinions, and see if we can find a good solution.

Sloy:
**This is a big PR, sorry for that. But it also an important one**

This fixes #116 applying the discussed solution.

Barista does a lot of magic in some methods via try/catch, for example to try multiple scenarios when trying to click a view, or a menu, or scrolling, etc.

The problem is that wrapping failing code in a try/catch avoids sending a failure to the test, but Espresso internally notifies its FailureHandler anyways. 

Normally this isn't an issue, unless you want to use a custom FailureHandler for other things. For example to take screenshots when a test fails.

## Solution
I created a `SpyFailureHandler` that captures all errors received and rethrows the exception. That way we can chain try/catch calls without notifying the global Espresso FailureHandler. If everything else fails, we can redirect the failure to the global handler. Generally we want to send the error from the simplest scenario for easier debugging (sometimes will be the first one, sometimes the last one).

We replace the FailureHandler in 2 ways:
- _The good:_ By using `ViewInteraction`'s method `.withFailureHandler(spyFailureHandler)` we use our own handler for a specific action.
- _The bad:_ In `DataInteraction` there isn't an equivalent method. I created the helper function `withFailureHandler()` which wraps the code, setting our custom handler before executing it and resetting the original later.

Most of the affected methods had to be rewritten almost entirely, so I wrote them in Kotlin.


Sloy:
I left the WIP label because I want to take a second look at the APIs, naming and retrocompatibility.

Sloy:
It can be tested by using this dependency: 
```gradle
androidTestCompile 'com.github.SchibstedSpain:Barista:better-error-handling-SNAPSHOT'
```

library/src/main/java/com/schibsted/spain/barista/BaristaAssertions.kt
```diff
@@ -0,0 +1,184 @@
...
+    fun assertThatBackButtonClosesTheApp() {
+        // Will launch an Exception if it closes the app
+        try {
+            onView(isRoot())
+                    .withFailureHandler(RethrowingFailureHandler())
+                    .perform(ViewActions.pressBack())
+            // One of our Activities is appearing on the screen :(
+            fail("The back button didn't close the app")
+        } catch (noActivityException: NoActivityResumedException) {
+            //Yey!
```
rocboronat: I propose just a "Yey! The back button closed the app!". It's more expressive.

library/src/main/java/com/schibsted/spain/barista/BaristaClickActions.kt
```diff
@@ -0,0 +1,84 @@
...
+import com.schibsted.spain.barista.custom.DisplayedMatchers.displayedAnd
+import com.schibsted.spain.barista.custom.NestedEnabledScrollToAction.nestedScrollToAction
+import com.schibsted.spain.barista.internal.failurehandler.SpyFailureHandler
+import com.schibsted.spain.barista.internal.failurehandler.description
+import com.schibsted.spain.barista.internal.resourceMatcher
+import org.hamcrest.Matcher
+import org.hamcrest.Matchers.allOf
+import org.hamcrest.Matchers.anyOf
+
+object BaristaClickActions {
```
rocboronat: Why `object`? That's the Kotlin way to do a Singleton, isn't it? And why it's important to be a singleton instead of just a class?

library/src/main/java/com/schibsted/spain/barista/BaristaListViewActions.kt
```diff
@@ -0,0 +1,94 @@
...
+import android.widget.AdapterView
+import com.schibsted.spain.barista.custom.PerformClickAction.clickUsingPerformClick
+import com.schibsted.spain.barista.internal.failurehandler.SpyFailureHandler
+import com.schibsted.spain.barista.internal.failurehandler.withFailureHandler
+import org.hamcrest.Matchers.*
+
+object BaristaListViewActions {
+
+    @JvmStatic
+    fun clickListViewItem(@IdRes listViewId: Int, vararg positions: Int) {
```
rocboronat: Not related with this PR... but I think that this method is useless. I wrote it because I thought it was an smart idea... but I can't think in a usecase for it. Just code that has to be understood and maintained.

library/src/main/java/com/schibsted/spain/barista/BaristaRecyclerViewActions.java
```diff
@@ -15,7 +14,7 @@
 
   public static void clickRecyclerViewItem(@IdRes int recyclerViewId, int... positions) {
     if (positions.length == 0) {
-      throw new BaristaException("positions cannot be empty");
+      throw new IllegalArgumentException("positions cannot be empty");
```
rocboronat: Previously, we wrote "Positions" with Capital letter. Let's maintain just one writing style.

Sloy:
I tried it at InfoJobs project and some tests fail, but I'm not sure how the errors relate to the changes. Let me investigate. But... I'm going on vacation!! \o/ So I won't be able to do it for some time.

library/src/main/java/com/schibsted/spain/barista/BaristaRecyclerViewActions.java
Sloy: Humm according to the diff it was lowercase before. But you're right, we should use capital letters for correctness.

library/src/main/java/com/schibsted/spain/barista/BaristaClickActions.kt
Sloy: It's the way of doing static methods in Kotlin, if I'm not wrong.

library/src/main/java/com/schibsted/spain/barista/BaristaClickActions.kt
```diff
@@ -43,14 +41,14 @@ object BaristaClickActions {
         try {
             try {
                 clickDisplayedView(viewMatcher, spyHandler)
-            } catch (exception: NoMatchingViewException) {
+            } catch (fistError: RuntimeException) {
```
SmasSive: typo! I don't think this kind of error is like a fist 👊 right?

Sloy: Haha should be!

library/src/main/java/com/schibsted/spain/barista/BaristaClickActions.kt
sch-cbavel: The class approach would be more verbose if all the members are static:

```
object BaristaClickActions {
   // <-- static members
}
```
--->

```
class BaristaClickActions {
   companion object {
        // <-- static members
   }
   // <-- instance members
}
```

In the end objects and companion objects act like static member containers.
The rationale about objects over statics is that the object concept allows you to extend other classes and implement interfaces while plain statics don't.

library/src/main/java/com/schibsted/spain/barista/BaristaScrollActions.kt
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @JvmStatic
+    fun safelyScrollTo(text: String) {
+        scrollWithMultipleAttempts(withText(text), failAtEnd = false)
+    }
+
+    private fun scrollWithMultipleAttempts(matcher: Matcher<View>, failAtEnd: Boolean) {
+        val spyFailureHandler = SpyFailureHandler()
+        for (i in 0..MAX_SCROLL_ATTEMPTS) {
```
sch-cbavel: Probably not too important, but this executes 51 times instead of 50, so not completely accurate :)
`1 .. MAX_SCROLL_ATTEMPTS` would fit better or `0 until MAX_SCROLL_ATTEMPTS`. But with the until you would have to change the i == MAX_SCROLL_ATTEMPTS

Sloy: Hum good point, I didn't think of that. The `i == MAX_SCROLL_ATTEMPTS` should have raised a warning in my head but it didn't 😅 

I'll go for the `1 .. MAX_SCROLL_ATTEMPTS`, it looks cleaner to me.

Thanks!!

Sloy:
Rebased with master.
I had to refactor the changes from https://github.com/SchibstedSpain/Barista/pull/136 too

Sloy:
Everything seems OK. Build is green and tests on our internal project work with this branch. 
Last chance for requesting changes....

library/src/main/java/com/schibsted/spain/barista/BaristaRecyclerViewActions.java
rocboronat: +1

rocboronat:
Noice! :·)

library/src/main/java/com/schibsted/spain/barista/BaristaRecyclerViewActions.java
Sloy: Oops! I thought I already did it!
Sorry @rocboronat and thanks for the reminder ❤️

rocboronat: No prob bro! 💋

Issue #100, opened 2017-06-20
Click() method name conflicts with Espresso's click() 


Sloy:
When writing tests with the auto-imports enabled, using the `click(id)` from Barista becomes a bit painful because Espresso has another static `click()` method, and Android Studio will automatically import the Espresso one first, forcing you to manually remove the import and use the Barista one.

This is an issue I've faced myself, and also I've been told by others (hi Maryna!).

My proposal is to rename the static `click(id)` methods to something like `clickOn(id)` so we don't conflict with Espresso. 
The old methods should probably be maintained to avoid breaking the public API, but at least deprecate them.

rocboronat:
Nice idea! Barista is meant to be human friendly. So, let's deprecate those Espresso-like `click()` methods.

Sloy:
Done by @alorma! 
It took some time @MarinaTov but it's finally fixed. Thanks for reporting it :)

PR #141, opened 2017-11-06
Rename click methods to clickOn

alorma:
Rename `click()` methods to `clickOn()` as issued on #100

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/BaristaClickActions.java
+++ b/library/src/main/java/com/schibsted/spain/barista/BaristaClickActions.java
@@ -29,6 +29,15 @@ public class BaristaClickActions {
   private static final ResourceTypeChecker RESOURCE_TYPE_CHECKER = new ResourceTypeChecker();
 
   // SingleClick Actions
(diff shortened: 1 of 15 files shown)


rocboronat:
If it's a pure IntelliJ refactor with zero creativity: +1 !

But I think that @Sloy is working on a huge refactor or something like thooot

library/src/main/java/com/schibsted/spain/barista/BaristaClickActions.java
```diff
@@ -29,6 +29,15 @@
...
   // SingleClick Actions
+  public static void clickOn(int id) {
+    performWithResource(id, ViewActions.click());
+  }
+
+  public static void clickOn(String text) {
+    performWithMatcher(withText(text), ViewActions.click());
+  }
+
+  // Deprecated SingleClick Actions
```
Sloy: You should use the deprecated annotation to tell consumers what's going on.
Also, if this goes into Barista 2.0 I wonder if we should deprecate or simply replace the method breaking the API.

Issue #86, opened 2017-06-01
Move internal classes to an internal package


Sloy:
We have many classes in Barista. Visible ones.

I think it would be a good idea to use packages to separate the public API and the internal classes. I've seen this in many libraries.
It means moving all classes intended for internal usage to an `internal` subpackage.

What do you think @rocboronat?

rocboronat:
I think that it will be great :·)

PR #143, opened 2017-11-08
Changed package structure

Sloy:
Fixes https://github.com/SchibstedSpain/Barista/issues/86

Now we have 3 public packages related to the 3 main feature categories of Barista and 1 internal package:
- Action
- Assertion
- Rule
- _Internal_

This is the structure now:
screen shot 2017-11-08 at 12 42 00

And this is when expanded:
screen shot 2017-11-08 at 12 42 22


rocboronat:
Great work! And thanks for the preview :-)

Issue #57, opened 2017-05-10
Avoid asking for an ID if there is only one ListView


rocboronat:

Sloy:
We should provide both methods. The current implementation of `BaristaListViewActions` is not that great. 

Brace yourselves, refactor is coming.

Sloy:
Should we also merge ListView methods with RecyclerView methods?
It sounds like something Barista would do.

Sloy:
This is being done in https://github.com/SchibstedSpain/Barista/pull/146

Issue #140, opened 2017-11-06
Deprecate unwanted APIs


Sloy:
Here's a list of APIs that we want to deprecate (either remove completely or replace them).

- [x] [replace] **ClickActions**: Rename click() method (https://github.com/SchibstedSpain/Barista/issues/100)
- [x] [replace] **Drawer actions/assertions**: We're using deprecated methods. We should explicitly let to select left or right drawer. Also, id should be optional in case there's only one drawer.
- [x] [replace] **ListViewActions**: The multiple positions parameters should be replaced with a single position parameter.
- [x] [replace] **RecyclerViewAction scrollTo**: This method can be confused with the normal scrollTo action.
- [x] [replace] **writeToAutoCompleteTextView**: Doesn't follow the Barista "simple method names" rule.
- [x] [replace] **writeToEditText**: Doesn't follow the Barista "simple method names" rule.
- [x] [remove] **CheckBoxClickActions**: They're just an alias to the normal click. Less APIs to maintain, happier developers.

PR #146, opened 2017-11-10
Merge RecyclerView and ListView actions into ListActions

Sloy:
Big news! With this PR I've merged together ListView and RecyclerView actions. No need to use different methods depending on implementation anymore. Yey! 💃 

This was part of https://github.com/SchibstedSpain/Barista/issues/140
Also fixes https://github.com/SchibstedSpain/Barista/issues/57

The key to this PR is in [BaristaListActions.kt](https://github.com/SchibstedSpain/Barista/blob/improved-list-actions/library/src/main/java/com/schibsted/spain/barista/action/BaristaListActions.kt). This class has 3 public methods for the 3 functionalities that we support on lists (always by position): 
- **clickListItem**([id,] position)
- **scrollListToPosition**([id,] position)
- **clickListItemChild**([id,] position, childId)

Also, view id is now optional when there's only one listview or recyclerview on the screen.

I wrote the code doing some TDD. I had to create a test activity allowing to generate dynamic UI with different kinds and numbers of lists. I hope I didn't miss any edge case.

Finally, this PR also removes the now obsolete BaristaListViewActions and BaristaRecyclerViewActions.

Happy testing!


library/src/main/java/com/schibsted/spain/barista/action/BaristaListActions.kt
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private fun performOnRecycler(id: Int?, failureHandler: FailureHandler, recyclerAction: ViewAction) {
+        onView(findRecyclerMatcher(id))
+                .withFailureHandler(failureHandler)
+                .perform(recyclerAction)
+    }
+
+    private fun performOnListView(failureHandler: FailureHandler, id: Int?, position: Int, listViewAction: ViewAction) {
+        withFailureHandler(failureHandler) {
```
sch-cbavel: Unrelated to this PR, but reading this wanted to comment:

What about changing:
`withFailureHandler(failureHandler) { ... }` to `failureHandler { ... }` or `failureHandler.apply { ... }` or `failureHandler.using { ... }`

So instead of a global function: `inline fun withFailureHandler(failureHandler: FailureHandler, scope: () -> Unit) {}` it would be `operator inline fun FailureHandler.apply(scope: () -> Unit) {}` (for example)

The idea behind that change is that you are repeating "failureHandler" twice when writing it

rocboronat:
Nice work, but I think you worked too much. Why not just using the old tests that checked the ListView and the RecyclerView Activities, but using the new methods? That's the easier approach to meet the need.

By the way, just a small _mimimi_: the title of the PR is not too descriptive :·P

Sloy:
@rocboronat because old tests didn't cover all scenarios (single list, multiple lists, no lists, different lists). And our ListView and RecyclerView actions were not the same, each one allowed different actions (one had scroll, the other one had clicking on a child). Covering all new cases in both tests would mean duplicating a lot of code.

> mi mimi mi mi mimimi mi mi mi mimimi :·P

You're right, any suggestion to explain it better in one line?

library/src/main/java/com/schibsted/spain/barista/action/BaristaListActions.kt
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private fun performMagicAction(@IdRes id: Int?, position: Int, recyclerAction: ViewAction, listViewAction: ViewAction) {
+        verifyOneSingleMatch(id)
+
+        val spyFailureHandler = SpyFailureHandler()
+        try {
+            performOnRecycler(id, spyFailureHandler, recyclerAction)
+        } catch (noRecyclerMatching: NoMatchingViewException) {
+            performOnListView(spyFailureHandler, id, position, listViewAction)
```
SmasSive: Just a mimimi, parameters order should be the same as `performOnRecycler` method, the first one has to be the `id`

library/src/main/java/com/schibsted/spain/barista/action/BaristaListActions.kt
Sloy: Sorry @sch-cbavel, I think you lost me.
I'm not seeing the difference between both approaches (apart from the syntax)

sch-cbavel: I meant that with this approach, since you are usually using **failureHandler** as the name of the variable, you are writing the name of the class twice (except for the case) :)

with**FailureHandler**(**failureHandler**) {
-->
**failureHandler**.apply {

Sure it is not a big deal, but reduces verbosity and its DRYer

Sloy: Hmm aren't `apply` and `with` type of methods usually intended for doing operations over the object itself?
In this case we're not using the failureHandler inside the code block, we're just using it internally to workaround Espresso.

IMO it might confuse readers. Not sure if it's just me.

rocboronat:
Oh @Sloy ! I didn't get that you added that features on both lists! WOW! 👏 👏 👏

library/src/main/java/com/schibsted/spain/barista/action/BaristaListActions.kt
sch-cbavel: Sorry. My fault. Wanted to put invoke, not apply:

```
class FailureHandler {
    operator fun invoke(callback: () -> Unit) {
        // use...
        try {
            callback()
        } finally {
            // restore
        }
    }
}

fun test() {
    val failureHandler = FailureHandler()

    failureHandler {
    }
    
    // or

    failureHandler.invoke {
    }
}
```

Other possible verbs:
```
failureHandler.temporally {
}
failureHandler.locally {
}
...
```

I'm pretty used to Kotlin DSLs and find this clean. But if you think it can confuse people. It is ok to leave as it is :) :+1:

Issue #133, opened 2017-10-05
Add pullToRefresh method without id


Sloy:
Currently the pull to refresh method (named `refresh()`) requires the layout id as a parameter. In most cases when there's only one SwipeRefreshLayout in screen there's no need to pass the id.

We can infer the view with a `isAssignableFrom(SwipeRefreshLayout.class)` matcher, so the id parameter wouldn't even be needed. We can still leave the the method with the parameter, just in case someone needs it.

rocboronat:
Noice!

EsteveAguilera:
This issue can be closed, it was fixed on #150

Sloy:
Yey!

PR #150, opened 2017-11-12
Add pull to refresh without id

EsteveAguilera:
Add a `refresh()` method without the need to add an `id` as parameter

Fixes the #133 with special thanks to @Sloy for the hint! 😜


Sloy:
You're on fire! 🔥 
Let me review everything from work tomorrow :)

If you're curious, you can take a look at https://github.com/SchibstedSpain/Barista/pull/146 to see a cool way to do the id magic by using kotlin's default parameter values.

EsteveAguilera:
You're right @Sloy! 😚 

I've converted the class to Kotlin and used the default parameter value

I've also renamed the `swipeRefreshId` parameter to `id`... This seems to be the convention on the rest of the project

Take the time you need to review all my changes! 
These are my first contributions, and I'm still not too familiar with the project...

EsteveAguilera:
🕺

PR #155, opened 2017-11-14
Give a handy way to fake the camera intents

rocboronat:
Right now, faking the response of a call to the Camera is not trivial.

Let's check how to do it with regular Espresso, and try to give a simpler approach with Barista.


alorma:
But... this is not generating a valid image on the Intent response :S

It will need something like ... create a Bitmap and store it on the file uri provided

rocboronat:
@alorma it was 4AM and the PR has the wip label 😏 Sounds like just sharing a wip 🐒

alorma:
Oppps.

I've created another PR with camera intent,BTW, based on this one

rocboronat:
That makes my work useful 😘 well done!

alorma:
Not at all @rocboronat I've used your base, and just improved it with intents :$

rocboronat:
Yas, and that's lovely 😘 I went to bed with work and I woke up with the work done 👌🏻 We're a great team 😎

PR #156, opened 2017-11-14
Add ImageView assertions for any or empty drawable

rocboronat:
We will need it when we don't know what should appear in an `ImageView`. This is directly related to #155 , where we will take a picture, and then we will just check that an `ImageView` has been filled with it.

In addition, this PR also adds an assertion to check that an ImageView doesn't have an assigned drawable.


library/src/main/java/com/schibsted/spain/barista/BaristaAssertions.java
```diff
@@ -162,6 +163,10 @@ public static void assertHint(@IdRes int id, String text) {
     onView(withId(id)).check(matches(withHint(text)));
   }
 
+  public static void assertDrawable(@IdRes int id) {
+    onView(withId(id)).check(matches(withDrawable(DrawableMatcher.Companion.getANY())));
```
rocboronat: That `getANY()` hurts my lovely new eyes 👀 but I don't know how to do it better

SmasSive: Hi @rocboronat !
You can annotate `DrawableMatcher` as this:

```kotlin
companion object {
    private val EMPTY = -1
    @JvmStatic
    @get:JvmName("any")
    val ANY = -2
  }
```
And then you can use it from java like this:
```kotlin
DrawableMatcher.any()
```

rocboronat: Woooow thanks for sharing! Now, I don't know if I like more the problem or the solution 😄 Let's wait for more feedback... BTW @SmasSive we need you at #159 !

rocboronat: @Sloy we need your feedback. I personally think that both solutions are not too pretty 😅

SmasSive: I think it's the price you have to pay to live with Java. You can still use only the annotation `@JvmStatic` to avoid the use of `Companion` but you will have to use `getANY()`.
Let's see what @Sloy has to say about it...

Sloy: I was trying to understand this PR yesterday afternoon, but wasn't a good moment for that :D
Please give me some time to finish some other things and I will be back 🤖

rocboronat: Take your time 😏😘

rocboronat: @Sloy not too much time, please... :·/

Sloy: Hi! Missed me? 💋  
Sorry 🙏 

**Here's my opinion:**
EMPTY and ANY look like internal constants that we should avoid exposing to the outside world. If we wanted to make them public we should use an enum or something more helpful.

**Here's my proposal:**
- Keep the constants privates to avoid leaking them to the outside.
- Make the DrawableMatcher constructor private.
- Provide public static factory methods in the companion object for the three kinds of matchers: `withId(drawableId)`, `any()`, `empty()`.

sample/src/androidTest/java/com/schibsted/spain/barista/sample/AssertionsTest.java
```diff
@@ -271,12 +271,22 @@ public void checkBackButton() {
   }
 
   @Test
-  public void checkSameDrawable() throws Exception {
+  public void checkDrawable_withId_withSameDrawable() throws Exception {
```
Sloy: Test names look confusing, it had to close my eyes really hard and think with all the power of my brain to understand what it means.

Here's a debatable proposal:
- checkDrawable_withId
- checkDrawable_withId_failure
- checkDrawable_withAnyDrawable
- checkDrawable_withAnyDrawable_failure

library/src/main/java/com/schibsted/spain/barista/BaristaAssertions.java
Sloy: Hi! Missed me? 💋  
Sorry 🙏 

**Here's my opinion:**
EMPTY and ANY look like internal constants that we should avoid exposing to the outside world. If we wanted to make them public we should use an enum or something more helpful.

**Here's my proposal:**
- Keep the constants privates to avoid leaking them to the outside.
- Make the DrawableMatcher constructor private.
- Provide public static factory methods in the companion object for the three kinds of matchers: `withId(drawableId)`, `any()`, `empty()`.

Sloy:
I rebased the branch with the latest master

Sloy:
NOTE: There is an important discussion on an "outdated" review about the companion constants.
GitHub outdated it because the class name changed, but the comments are still valid.

sample/src/androidTest/java/com/schibsted/spain/barista/sample/AssertionsTest.java
rocboronat: Mmmm... it's not the moment to debate 😂 let's finish all this crap

rocboronat:
Wep! While doing the requested changes I noticed that it's really easy to also add an assertion to check that an ImageView has not an assigned drawable. So, I added it to the public API.

README.md
```diff
@@ -132,8 +132,12 @@ assertRecyclerViewItemCount(R.id.recycler, 10);
 // And another tricky feature
 assertThatBackButtonClosesTheApp();
 
-// Is this ImageView showing this drawable?
+// Is this ImageView showing a drawable?
+assertDrawable(R.id.image_view);
 assertDrawable(R.id.image_view, R.drawable.ic_barista);
+
+// ...or not?
+assertWithoutDrawable(R.id.image_view);
```
rocboronat: @Sloy we should move this entry to the ones that have the style "blablabla / ...or not?". It's the moment to do it? Can we do it later? (My vote for later)

library/src/main/java/com/schibsted/spain/barista/internal/matcher/DrawableMatcher.kt
```diff
@@ -18,6 +18,9 @@ class DrawableMatcher(private val expectedId: Int) : TypeSafeMatcher<View>(View:
   companion object {
     private val EMPTY = -1
     private val ANY = -2
+
+    fun empty() = EMPTY
+    fun any() = ANY
```
rocboronat: After doing what you requested, I think this code doesn't add value. I know that they are accessors to private constants but it doesn't worth the price IMHO. Just sayin'!

README.md
Sloy: Later is good for me

library/src/main/java/com/schibsted/spain/barista/assertion/BaristaImageViewAssertions.kt
```diff
@@ -2,16 +2,25 @@ package com.schibsted.spain.barista.assertion
...
 object BaristaImageViewAssertions {
 
     @JvmStatic
     fun assertDrawable(@IdRes id: Int, @DrawableRes drawable: Int) {
-        Espresso.onView(ViewMatchers.withId(id)).check(ViewAssertions.matches(DrawableMatchers.withDrawable(drawable)))
+        onView(withId(id)).check(matches(DrawableMatchers.withDrawable(drawable)))
     }
 
+    @JvmStatic
+    fun assertDrawable(@IdRes id: Int) {
```
Sloy: Sorry for the late comment:
Now that we have multiple drawable assertions, I think the naming should be changed to explain the difference between each method:
- assertHasDrawable(id, drawable)
- assertHasAnyDrawable(id)
- assertHasNoDrawable(id)

sample/src/androidTest/java/com/schibsted/spain/barista/sample/introduction/IntroducingBaristaAssertions.java
```diff
@@ -75,7 +76,8 @@ public void letsIntroduceBarista() {
     // And another tricky feature
     assertThatBackButtonClosesTheApp();
 
-    // Is this ImageView showing this drawable?
-    assertDrawable(R.id.image_view, R.drawable.ic_barista);
+    // Is this ImageView showing a drawable?
+    assertHasAnyDrawable(R.id.image_view);
+    BaristaImageViewAssertions.assertHasDrawable(R.id.image_view, R.drawable.ic_barista);
```
rocboronat: UH!

rocboronat: Oh, this file will be deleted by @Sloy. Ignoring it in 3, 2, 1...

rocboronat:
And that's all folks! I think it's ready to merge, @Sloy @SmasSive

rocboronat:
BTW, ran all tests and they give green 💚

library/src/main/java/com/schibsted/spain/barista/internal/matcher/DrawableMatcher.kt
Sloy: You're right. Because that's not what I meant with factory methods 😂 

I mean:
```kotlin
companion object {
  private EMPTY=...
  private ANY=....
  
  fun any() = DrawableMatcher(ANY)
  fun empty() = DrawableMatcher(EMPTY)
  fun withId(@Idres id: Int) = DrawableMatcher(id)
```

It's the equivalent to the typical Java static methods to create the object. Think of the `createIntent(....)` pattern.

SmasSive: @rocboronat I like what @Sloy suggests...
@Sloy you owe me one beer 🍺

rocboronat: I like it too <3 Feel free to add code in your examples, to avoid misunderstanding your explanations. It will save time and effort. BTW, feel free to commit your suggestions to the PR. I will not discuss it, so let's simplify the development process.

Issue #120, opened 2017-08-12
assertText(R.id.widget, "expected text")


estefafdez:
There is no way to get the text on a textView and compare with a string parameter to check if the text is the same or not.

rocboronat:
That's not the way to use Espresso nor Barista. Instead of getting the text of the `TextView`, just use Barista's `assertDisplayed(yourExpectedText)`

rocboronat:
Hello @estefafdez! Does my answer solve the question?

estefafdez:
Hello, yes, but if I want to get the text from the textView because the text was modified on different activities, is there any way to get it? Only with the assertDisplayed?

rocboronat:
Yes, the idea is that Espresso doesn't encourage us to get the data from a widget. So, the Espresso way to check that this text is being shown is to use the `assertDisplayed(something)` method.

If you need to check that a given text is shown in a given `TextView`, we could write a `assertText(R.id.widget, "expected text")` method, but if you can go on with the simple `assertDisplayed()` one... :·D

estefafdez:
The assertText method sounds great, I think it could be helpful for people like me, used to work with tools such as Appium or Selenium on Automation Testing. Thank you!

rocboronat:
Ok! Let me add to the backlog :·)

estefafdez:
Perfect, thanks, a lot!

EsteveAguilera:
This issue can be closed, it was fixed with the merge of #165!

PR #157, opened 2017-11-14
 Add assertText assertion to check that a given text is shown in a given View

EsteveAguilera:
Add assertion 
`assertText(R.id.widget, "expected text")`

This assertion was requested on issue #120

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaAssertions.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaAssertions.kt
@@ -40,6 +40,11 @@ object BaristaAssertions {
         assertDisplayed(withText(text))
     }
 
+    @JvmStatic
+    fun assertDisplayed(@IdRes resId: Int, text: String) {
+        onView(resId.resourceMatcher()).check(matches(withText(text)));
+    }
+
     private fun assertDisplayed(matcher: Matcher<View>) {
         val spyFailureHandler = SpyFailureHandler()
         try {
@@ -200,10 +205,5 @@ object BaristaAssertions {
     @JvmStatic
     fun assertNotFocused(text: String) {
         onView(withText(text)).check(matches(not(hasFocus())))
-    }
-
-    @JvmStatic
-    fun assertText(@IdRes resId: Int, text: String) {
-        onView(resId.resourceMatcher()).check(matches(withText(text)));
     }
 }
--- a/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaAssertions.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaAssertions.kt
@@ -201,4 +201,9 @@ object BaristaAssertions {
     fun assertNotFocused(text: String) {
         onView(withText(text)).check(matches(not(hasFocus())))
     }
+
+    @JvmStatic
+    fun assertText(@IdRes resId: Int, text: String) {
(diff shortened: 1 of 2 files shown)


sample/src/androidTest/java/com/schibsted/spain/barista/sample/AssertionsTest.java
```diff
@@ -305,4 +311,9 @@ public void checkViewHasNotFocus() throws Exception {
     assertNotFocused(R.string.edittext_with_no_focus);
     assertNotFocused("EditText with no focus");
   }
+
+  @Test
+  public void checkExpectedText() throws Exception {
```
rocboronat: A small comment there: can you move this test near the other test methods that check for texts? Just to have them grouped 🙌

Sloy:
Hmm I wonder: if we use [assertDisplayed(text)`, why changing the naming instead of using `assertDisplayed(id, text)`? Do we need to make a distinction?

sample/src/androidTest/java/com/schibsted/spain/barista/sample/AssertionsTest.java
EsteveAguilera: Of course!

rocboronat:
+1 to @Sloy. `assertDisplayed` sounds like the method that checks for everything that it displayed, and we could offer lots of it.

EsteveAguilera:
Ok, two owners is a KO win!

rocboronat:
I proposed the bad name at #120, so... one of the owners is the guilty 🍶

EsteveAguilera:
Ooops, I just realized I've been working directly on my **barista2** branch instead of a new one for the PR... sorry about that! 😳 

The last commit also moves the test @rocboronat, I hope it's on a better place now 😅

rocboronat:
That's the Fewlaps performance! 💪

rocboronat:
That's the Android Studio autoformatter performance! 💪

library/src/main/java/com/schibsted/spain/barista/assertion/BaristaAssertions.kt
```diff
@@ -40,6 +40,11 @@ object BaristaAssertions {
         assertDisplayed(withText(text))
     }
 
+    @JvmStatic
+    fun assertDisplayed(@IdRes resId: Int, text: String) {
+        onView(resId.resourceMatcher()).check(matches(withText(text)));
```
Sloy: Watch out! Here you don't want to use our magic (I think). 
`resourceMatcher()` is intended to give you the matcher for an id **or** a string resource.

That means you can use `assertDisplayed(R.string.hello, "hello")`. I think it's best that we limit the matcher to `withId(id)`, so users won't get unexpected results if they use the API wrong.

rocboronat: Uuuuuuh it's the **EDD** time!

sample/src/androidTest/java/com/schibsted/spain/barista/sample/AssertionsTest.java
```diff
@@ -64,6 +64,11 @@ public void checkVisible_withRepeatedViews() throws Exception {
   }
 
   @Test
+  public void checkExpectedText() throws Exception {
```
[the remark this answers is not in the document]
EsteveAguilera: 👍

EsteveAguilera:
😊

PR #159, opened 2017-11-14
Fake camera calls intents

alorma:
Based on https://github.com/SchibstedSpain/Barista/pull/155

I've used a custom `matcher<Intent>` to retrieve the uri form `MediaStore.EXTRA_OUTPUT`, and creates a fake Bitmap on it


alorma:
@Sloy could you apply the codestyle? As it uses the internal Schisbsted one... not easy for me

rocboronat:
Yep @alorma, is this a wip? I mean, are you going to add the magic to Barista, or the idea is to merge this PR as is?

alorma:
@rocboronat What do you mean?

I've added it to library.

You mean add it to Readme.md?

rocboronat:
My idea was to offer a Barista simple way of faking the camera with less code than possible. Right now, if you look at https://github.com/SchibstedSpain/Barista/pull/159/files#diff-d5f3e17e30aa7c9ceaa925bdedec6c37, you will see that the user has to use another rule, write two lines of Espresso _mimimi_, and four of Android. 

The expectation was to just call a Barista method to do all the magic. For example, `BaristaIntents.fakeCamera()`, and enabling the use of the `BaristaRule`, ignoring the `IntentsTestRule`. You know, something catchy to remember and ultra easy to use.

alorma:
In fact, user not needs to use `IntentsTestRule` at all. Just remember to call `Intents.init()` & `Intents.release()`.

I don't think is a good idea to fake everything, as it needs to actions:

`intending` -> For "mock" intents
... (test code) ...

`intended` -> For verify intents

IMHO is good enough for *Barista* to provide a cameraMatcher, that automatically generates a picture, so everything works as expected, without changing production code

Sloy:
Why not call Intents.init() and Intents.release() inside the Barista method?

It should work if we don't do any operations outside it. Right?

_P.S.: I haven't read the PR yet._

alorma:
No @Sloy 

`init()` has to be called BEFORE `intending`
`release()` has to be called AFTER `intended()`

So it cannot be done inside a single method, unless we apply some kind of block extension (for kotlin):

```
cameraFake() {
 ... test code ...
}
```

And it doesn't looks fine for me.

In tis PR i just provide a simple matcher, that does a fake camera, creating an image on the URI provided

rocboronat:
@alorma providing a matcher is good, but giving the whole magic is better. Barista is an Espresso wrapper that makes using Espresso a happy game. If we say "you can do it with regular Espresso" instead of giving a great API, why are we developing Barista?

My vision of usage:
```java
@Test
public void takePictureAndShowIt() {
    BaristaIntents.fakeCamera()
    click(R.id.take_picture);
    assertDrawable(R.id.image_view);
}
```

alorma:
You can't do that @rocboronat

Because the `intendend()` is like `verify()` you can't set it above the "when" part, can you?

rocboronat:
We are doing it at https://github.com/SchibstedSpain/Barista/pull/159/files#diff-d5f3e17e30aa7c9ceaa925bdedec6c37, isn't it?

alorma:
And AFAIK, it's mandatory to setup `intended()` or it will opens the camera app.

Let me check

alorma:
ok... i'm missing something then u_u

i will check

rocboronat:
Thanks @alorma!

alorma:
Done @rocboronat you were right

rocboronat:
Great! Thanks for the work :·) Now, let's talk about the name of the method. Imagine that you don't know now nothing about the Espresso Intents framework. If you want to fake the camera, what's the best name to discover that method? I know that my proposal, `fakeCamera` is not the best one, but `captureIntent` doesn't sound like something related to the camera...

Guys, the naming discussion arrived :·) @Sloy @sergiandreplace @EsteveAguilera ... feel free to invite more people to the party! (bring cookies)

library/src/main/java/com/schibsted/spain/barista/intents/BaristaIntents.java
```diff
@@ -0,0 +1,31 @@
...
+import android.content.Intent;
+import android.os.Bundle;
+
+import static android.support.test.espresso.intent.Intents.intending;
+import static com.schibsted.spain.barista.intents.BaristaIntentMatchers.captureImage;
+
+public class BaristaIntents {
+
+  public static void captureIntent() {
+    captureIntent(100, 100);
```
rocboronat: Let's extract those magic numbers to `DEFAULT_SIZE` or something like that

alorma:
Maybe you know nothing (Jon Snow) of Esspresso intents...

But you know you had a Intent in your code, and you know it's the MediaProvider.Action_Capture intent...

rocboronat:
For me, capturing an intent sounds more like something related to the Mockito Captors than something related to faking the Android camera

EsteveAguilera:
+1 to @rocboronat ... I see more descriptive **fakeCamera**, but I agree that it's not the best naming...

IMO: For test purposes I don't need to know nothing about intents, I just want to launch the camera and take a photo... so my proposals:
`captureFakePicture `
`captureFakePhoto `

rocboronat:
`mockAndroidCamera`?

alorma:
Done @EsteveAguilera @rocboronat

rocboronat:
Thanks @alorma ! By the way, let's wait for other collaborators opinions because naming is the harder part of writing a library... And we have written the feature in less than 12 natural hours, including a night. We can wait a little for the best name ever seen 😂

aballano:
`barista.takeHipsterSelfie()` :trollface: 

Otherwise `mockAndroidCamera` or `mockCamera` looks good :)

(my two cents)

sample/build.gradle
```diff
@@ -22,9 +23,13 @@ android {
 }
 
 dependencies {
+  compile fileTree(dir: 'libs', include: ['*.jar'])
```
SmasSive: You don't need this line, do you?

SmasSive:
About naming, I do like `mockAndroidCamera()` but still sounds a little too much techie, if this is intended ok but if not, I'd suggest `givenACamera()`, `testableCamera`, `fakeCamera()`... Something not related to Android, mockito or some techie concept.

PS: I'm the worst in naming, feel free to ignore me.

sample/build.gradle
alorma: uhm.... i guess it was automatically changed, because i don't remember adding or updating it

alorma: Done @SmasSive

sergiandreplace:
and why not just takePicture()? I think it makes easier to read.

rocboronat:
@sergiandreplace the thing is that this magic line will not take a picture. It will prepare the test environment to fake the calls of the production app to the camera. Instead of opening the camera (and lose the test control), Espresso will receive that intent, create a file with an image in the Uri the app requested, and return the control to the app calling the common `onActivityResult()`. It's like doing the work of the camera, but without doing it 📷

yeradis:
Hello everyone,

As @rocboronat pointed out, the camera here does nothing more than to provide a name, cause its not really calling the camera, just faking the call and the result, but a specific one: placing a image at the specific uri.

my suggestion for the name: `fakeCameraResult`or `mockCameraResult`

regards

alorma:
`fakeCameraResult()`

Everyone is ok with that name?

@Sloy 
@SmasSive

rocboronat:
For me is the best approach :·) But let's wait for @Sloy, because it's the most commited to the project :·D

alorma:
As described in: https://developer.android.com/training/camera/photobasics.html

When the Intent doesn't contains the URI path for photo, it returns a thumbnail on the bundle extras.

Should Barista do the same? Return the bitmap (small one) on the bundle itself?

@rocboronat 
@Sloy 
@SmasSive

rocboronat:
@Sloy hey, just a reminder about reviewing this PR.

EsteveAguilera:
LGTM 😊

EsteveAguilera:
@alorma it seems that the PR now has conflicts... 
Can you fix them to make things easy for @Sloy?

alorma:
Done @EsteveAguilera

rocboronat:
@alorma the commit fails at Travis 😕

alorma:
Fixed @rocboronat

rocboronat:
Noice 👍

EsteveAguilera:
LGTM 👍

rocboronat:
@Sloy 🙏

rocboronat:
:tada:

Issue #131, opened 2017-09-29
Make the FlakyTestRule independent from ActivityTestRule


Sloy:
As noted by @CristianGM, our `FlakyTestRule` extends `ActivityTestRule` but there's no reason to do it. It would work as an independent rule too.

It's this way because of historic reasons of how we found a solution to the problem, but no longer needed.

Thanks @CristianGM for the heads up! 
_(and next time maybe open an issue or PR 😛)_

Sloy:
For retrocompatibility I think we should deprecate the current one but leave it as it is, and create a new simpler one. They can probably reuse some code though.

SmasSive:
I agree @sloy!

PR #160, opened 2017-11-14
 Create FlakyTestRule independent from ActivityTestRule

Sloy:
Solves #131

This change decouples FlakyTestRule from ActivityTestRule. We prefer composition over inheritance for rules.

I left the FlakyActivityTestRule as deprecated. Even though we're moving to Barista 2.0 with breaking changes, this one has a really low cost to maintain. We'll remove it in the future, reducing the migration impact for Barista 2.0.

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/rule/BaristaRule.java
+++ b/library/src/main/java/com/schibsted/spain/barista/rule/BaristaRule.java
@@ -2,10 +2,11 @@ package com.schibsted.spain.barista.rule;
 
 import android.app.Activity;
 import android.content.Intent;
+import android.support.test.rule.ActivityTestRule;
 import com.schibsted.spain.barista.rule.cleardata.ClearDatabaseRule;
 import com.schibsted.spain.barista.rule.cleardata.ClearFilesRule;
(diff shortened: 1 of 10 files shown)


library/src/main/java/com/schibsted/spain/barista/rule/BaristaRule.java
```diff
@@ -24,21 +25,26 @@
...
+        LAUNCH_ACTIVITY_AUTOMATICALLY);
   }
 
   @Override
   public Statement apply(Statement base, Description description) {
-    return RuleChain.outerRule(activityTestRule)
+    return RuleChain.outerRule(flakyTestRule)
+        // ↓ All rules below flakyTestRule will be repeated
+        .around(activityTestRule)
+        // ↓ All rules below activityTestRule will execute before launching the activity
```
rocboronat: Nice comments! 💘

Sloy: They're for future-me. Today I had to make some drawings to remember how RuleChain order works 😅

rocboronat: I did the same yesterday 😢

Issue #151, opened 2017-11-13
Update the README when merging Barista 2


rocboronat:
Let's wait until the branch is merged... :·)

Sloy:
Yeh, the readme has to be completely updated for Barista 2.0. 
I'm accumulating everything until I finish the public API refactoring.

But thanks for the issue, that way we won't forget :)

rocboronat:
Let's change the shape of the issue, then :·D

[ ] Add the new focus assertions to the README.md

Sloy:
Updated at https://github.com/SchibstedSpain/Barista/pull/163

Sloy:
Done!

PR #161, opened 2017-11-14
Rename and remove some APIs

Sloy:
Fixes issue https://github.com/SchibstedSpain/Barista/issues/140

Because we're releasing a new major version with breaking API changes, we want to cleanup some of our APIs:

- Deleted CheckBoxActions because they were just an alias for clickOn(id).
- Simplify writeToEditText method name to make it more Barista-style.
- Simplify writeToAutoCompleteTextView method name to make it more Barista-style.
- Drawer actions and assertions now have optional id and gravity parameters.
- Drawer actions implemented with Espresso's non-deprecated methods.


library/src/main/java/com/schibsted/spain/barista/action/BaristaCheckBoxActions.java
```diff
@@ -1,16 +0,0 @@
-package com.schibsted.spain.barista.action;
-
-import android.support.annotation.IdRes;
-
-import static com.schibsted.spain.barista.action.BaristaClickActions.clickOn;
-
-public class BaristaCheckBoxActions {
-
-  public static void clickCheckBoxItem(@IdRes int checkBoxId) {
```
rocboronat: I'm not sure if deleting this method is savvy. As a Barista developer, you know that it's just a shortcut to a click, but as we offer lots of methods related to widgets, some one could search for `BaristaCheckBox` to find what he can do with checkboxes. Or it can also give more information about what's a given id. I worked on projects where you can't know what's an just looking at it, because some times we don't give great names, and having a `clickCheckBoxItem(R.id.list)` can give more information about what will be happening on the test. And... it's not too much code :·P

library/src/main/java/com/schibsted/spain/barista/action/BaristaEditTextActions.kt
```diff
@@ -10,8 +10,8 @@ import com.schibsted.spain.barista.internal.matcher.DisplayedMatchers.displayedA
 object BaristaEditTextActions {
 
     @JvmStatic
-    fun writeToEditText(@IdRes id: Int, text: String) {
-        safelyScrollTo(id)
-        onView(displayedAnd(withId(id))).perform(replaceText(text))
+    fun writeTo(@IdRes editTextId: Int, text: String) {
```
rocboronat: +1 to the new param name :·)

library/src/main/java/com/schibsted/spain/barista/internal/viewaction/AutoCompleteViewActions.java
```diff
@@ -16,7 +16,7 @@
 import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
 import static org.hamcrest.Matchers.allOf;
 
-public class AutocompleteViewActions {
+public class AutoCompleteViewActions {
 
   public static ViewAction replaceAutocomplete(@Nonnull String stringToBeSet) {
```
rocboronat: It the class has been renamed to AutoComplete, the method should too, to keep the naming style.

sample/src/androidTest/java/com/schibsted/spain/barista/sample/NavigationDrawerActivityTest.java
```diff
@@ -33,4 +38,26 @@ public void openAndCloseDrawer() {
...
+    assertDisplayed("menu item");
+    assertDrawerIsOpen();
+
+    closeDrawer();
+    assertNotDisplayed("menu item");
+    assertDrawerIsClosed();
+  }
+
+  @Test
+  public void openAndCloseDrawerAtRight() {
```
rocboronat: Suggestion: `openAndCloseDrawerAtEnd`. In addition, what happens if you use `GravityCompat.RIGHT`? Do you think we need to test it? If no, go on :·D

sample/src/androidTest/java/com/schibsted/spain/barista/sample/introduction/IntroducingBaristaAssertions.java
```diff
@@ -62,8 +62,8 @@ public void letsIntroduceBarista() {
     assertUnchecked(R.id.unchecked_checkbox);
 
     // What's the state of the Drawer
-    assertDrawerIsOpen(R.id.drawer);
-    assertDrawerIsClosed(R.id.drawer);
+    assertDrawerIsOpen();
```
rocboronat: I think we can also give the methods that check it by id or by gravity, just for the marketing.

library/src/main/java/com/schibsted/spain/barista/action/BaristaCheckBoxActions.java
Sloy: Meeeeeeh I'm not convinced. You could give the same argument for other widgets like Button or Switch. And it's a new API to learn.

I could see the benefit of having dedicated methods for checking/unchecking a checkbox, but I can't see any benefit on having two different methods for the same thing 🤔

library/src/main/java/com/schibsted/spain/barista/internal/viewaction/AutoCompleteViewActions.java
Sloy: True! Ma' mistake

sample/src/androidTest/java/com/schibsted/spain/barista/sample/NavigationDrawerActivityTest.java
Sloy: Yeh... I had doubts about it, to me "AtEnd" sounds like it's referring to a time dimension, not gravity. How about `openAndCloseDrawerAtEndGravity`?

About the second topic, you should expect it to work on the same gravity that your layout has. If you use `right` in your layouts, use `Gravity.RIGHT`. If you use `end`, use `GravityCompat.END`. It's not important which one you use, as long as you use the same in your app. Don't see a need for a special test. But thanks for raising the question! We could have missed a special case.

sample/src/androidTest/java/com/schibsted/spain/barista/sample/introduction/IntroducingBaristaAssertions.java
Sloy: Does this count as a readme? Because in that case I will update it in https://github.com/SchibstedSpain/Barista/issues/151

library/src/main/java/com/schibsted/spain/barista/action/BaristaCheckBoxActions.java
rocboronat: If I can't convince you it means that it doesn't have sense. Let's remove those methods! 🎉 🎈 😂

sample/src/androidTest/java/com/schibsted/spain/barista/sample/NavigationDrawerActivityTest.java
rocboronat: Noice!

sample/src/androidTest/java/com/schibsted/spain/barista/sample/introduction/IntroducingBaristaAssertions.java
rocboronat: Well, that method is what we used to make a copy paste to update the `README.md` 👅

Issue #132, opened 2017-10-05
Rename the concept of "Action"


Sloy:
## The problem
It's sometimes confusing, even for myself, the mix between _**Espresso actions**_ (as in a `ViewAction` that you must pass to the `perform()` method) and a _**Barista action**_ (as in a static void method that executes some view action underneath).

In Barista we have both, some custom view actions (like the **NestedScrollAction**) and a lot of static methods inside classes with the Action suffix (like **BaristaClickActions**).

## Proposal
I'm thinking about renaming the latter concept to something else, like "Interaction" or similar, to easily distinguish between ViewActions and Barista methods. Interaction is the best name I could find, but I'm open to suggestions.

## Trade-off
The trade-off is that this change would completely break the API contract, so it should be added in a 2.0.0 release at least. Even with that I'm not sure if it's too much of an API change.

However we could leave all the existing methods the current classes as deprecated aliases, and remove them in a future version, to ease the transition a bit.

**Please I want your opinion. What do everyone else thinks about this?**

sergiandreplace:
Interaction looks fine. Other synonims are act, maneuver, operation, procedure, step, activity (but this one will be maybe even more confusing), operation, reaction. All these according to thesaurus dictionary.

SmasSive:
I am as bad as you for finding according names so, Interaction looks fine for me.
Regarding breaking the API, I would vote for deprecate the current methods and remove them in further versions.

rubenaranamorera:
Agree with @SmasSive 
Interactions looks ok, and I think the deprecation of current methods can be the best solution for this.

rocboronat:
Great idea! @Sloy , in my humble opinion, gogogo with a 2.0 version that breaks all the API. Explain it at the releases tab, aiming people to do a search and replace, and no problem. As the idea is great and the solution is simple (just changing `Action` by `Interaction` on tests), choosing the easiest to maintain solution is the best for me. If we deprecate the code, we will have a lot of "shortcut" methods... and we will have this discussion again in the future. So... for me, let's do it now.

PR #162, opened 2017-11-14
Rename Barista Actions to Interactions

Sloy:
As decided in https://github.com/SchibstedSpain/Barista/issues/132 we're renaming our "action" classes to "interaction" to avoid confussion with Espresso's view actions.

Barista Interactions are method's that perform an UI interaction upon calling.
On the other hand, Espresso ViewActions are objects that get executed by some other component.

In Barista we have both, since some of our Interactions are implemented with custom ViewActions.

Will be hard getting used to the new naming, but hopefully will make our lives easier in the long term.


rocboronat:
bubbub

PR #163, opened 2017-11-15
Readme for 2.0

Sloy:
We're almost there! I updated the readme with the recent changes and wrote a migration guide for 2.0 with detailed breaking changes from 1.x.

In the readme, I've put the download section at the top to make it easier to copy&paste.

I did it incrementally, so I might have missed something. Please take a look:
- [Migration guide](https://github.com/SchibstedSpain/Barista/blob/readme2/MIGRATION-2.0.md)
- [New readme](https://github.com/SchibstedSpain/Barista/blob/readme2/README.md)

### DIFF
--- a/MIGRATION-2.0.md
+++ b/MIGRATION-2.0.md
@@ -30,7 +30,7 @@ com.schibsted.spain.barista.rule
 - `ListView` or `RecyclerView` id is optional when only one instance of `ListView` or `RecyclerView` is displayed in the hierarchy. ([#146](https://github.com/SchibstedSpain/Barista/pull/146)
 - Drawer id on drawer actions and assertions is optional when only one DrawerLayout is present in the hierarchy. ([#161](https://github.com/SchibstedSpain/Barista/pull/161))
 - Drawer actions and assertions allow using non-default gravity with `openDrawerWithGravity()`, `closeDrawerWithGravity()`, `assertDrawerIsOpenWithGravity()` and `assertDrawerIsClosedWithGravity()`. ([#161](https://github.com/SchibstedSpain/Barista/pull/161))
-- `SwipeRefreshLayout` id is not optional in `refresh()` method when only one `SwipeRefreshLayout` is present in the hierarchy. ([#150](https://github.com/SchibstedSpain/Barista/pull/150))
+- `SwipeRefreshLayout` id is now optional in `refresh()` method when only one `SwipeRefreshLayout` is present in the hierarchy. ([#150](https://github.com/SchibstedSpain/Barista/pull/150))
 - Added `assertFocused()` and `assertNotFocused()` methods.
 ([#157](https://github.com/SchibstedSpain/Barista/pull/157))
 - Added `assertDisplayed(id, text)` to check some text in a specific view. ([#150](https://github.com/SchibstedSpain/Barista/pull/150))
--- a/MIGRATION-2.0.md
+++ b/MIGRATION-2.0.md
@@ -1,0 +1,43 @@
+# Migration guide from Barista 1.x to 2.0
+
+You just updated Barista dependency from 1.x to 2.0 and everything is in red?
+We did some breaking changes to improve our public API, and that's why we increased our major version.
+
+But fear not, here are a list of breaking changes and how to solve them:
+
+## Packages and classes
+We changed our Action classes to Interaction, to avoid confusion with Espresso ViewAction. ([#132](https://github.com/SchibstedSpain/Barista/issues/132))
+
+We moved most classes to different packages for better code organisation. You will find your missing imports in one of these public packages:
+```
+com.schibsted.spain.barista.assertion
+com.schibsted.spain.barista.interaction
+com.schibsted.spain.barista.rule
+```
+
+## Changed methods
+- `click()` methods have been renamed to `clickOn()`. ([#100](https://github.com/SchibstedSpain/Barista/issues/100))
+- `writeToEditText()` method has been renamed to `writeTo()`. ([#140](https://github.com/SchibstedSpain/Barista/issues/140))
+- `writeToAutoCompleteTextView()` method has been renamed to `writeToAutoComplete()`. ([#140](https://github.com/SchibstedSpain/Barista/issues/140))
+- `clickCheckBoxItem()` methods have been deleted. Use `clickOn()` instead. ([#140](https://github.com/SchibstedSpain/Barista/issues/140))
+- `ListView` and `RecyclerView` actions have been merged into `ListActions` with these common methods ([#146](https://github.com/SchibstedSpain/Barista/issues/146)):
+  - `clickListItem(listId, position)`
+  - `scrollListToPosition(listId, position)`
+  - `clickListItemChild(listId, position, childId)`
+
+## New features
+- `ListView` and `RecyclerView` methods are the same. No need to use different method depending on the implementation. ([#146](https://github.com/SchibstedSpain/Barista/pull/146))
+- `ListView` or `RecyclerView` id is optional when only one instance of `ListView` or `RecyclerView` is displayed in the hierarchy. ([#146](https://github.com/SchibstedSpain/Barista/pull/146)
+- Drawer id on drawer actions and assertions is optional when only one DrawerLayout is present in the hierarchy. ([#161](https://github.com/SchibstedSpain/Barista/pull/161))
+- Drawer actions and assertions allow using non-default gravity with `openDrawerWithGravity()`, `closeDrawerWithGravity()`, `assertDrawerIsOpenWithGravity()` and `assertDrawerIsClosedWithGravity()`. ([#161](https://github.com/SchibstedSpain/Barista/pull/161))
+- `SwipeRefreshLayout` id is not optional in `refresh()` method when only one `SwipeRefreshLayout` is present in the hierarchy. ([#150](https://github.com/SchibstedSpain/Barista/pull/150))
+- Added `assertFocused()` and `assertNotFocused()` methods.
+([#157](https://github.com/SchibstedSpain/Barista/pull/157))
+- Added `assertDisplayed(id, text)` to check some text in a specific view. ([#150](https://github.com/SchibstedSpain/Barista/pull/150))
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -135,25 +163,34 @@ assertThatBackButtonClosesTheApp();
...
+// Delete all tables from all the app's SQLite Databases
+@Rule public ClearDatabaseRule clearDatabaseRule = new ClearDatabaseRule();
+
+// Delete all files in getFilesDir() and getCacheDir()
+@Rule public ClearFilesRule clearFilesRule = new ClearFilesRule();
 ```
 
-## Dealing with Flaky tests
+
+### Dealing with Flaky tests
```
Sloy: The FlakyActivityTestRule is deprecated in favor of FlakyTestRule+ActivityTestRule. But when writing the readme, it seems easier to just use FlakyActivityTestRule.

Should we un-deprecate it? Or should I write in the readme the best way to use the new one with a RuleChain? I'm having second thoughts.

README.md
```diff
@@ -6,13 +6,34 @@
 
 ![Barista logo](art/barista-logo.png)
```
Sloy: Maybe we should put a quick table of contents as in https://github.com/SchibstedSpain/Parallax-Layer-Layout

README.md
rocboronat: I think that the best thing we can do is to just aim everyone to use the BaristaRule. Actually, it's what I do in all projects I work and what I recommend to everyone. We could just change the way we explain it at the readme... Right now, the most explained rule is the Flaky one, but just because it was the first one that we written. Now, we can introduce the BaristaRule first, and then explain the Flaky rule and the other ones in another file. Well! But first, let's merge this PR :·D

README.md
rocboronat: Personally, I never used those quick tables... I think that they break what I expect from a readme. I think we should focus in making the readme shorter, moving the really technical things to other files. Then, we wouldn't need that quick table 😅

MIGRATION-2.0.md
```diff
@@ -0,0 +1,43 @@
...
+- `writeToAutoCompleteTextView()` method has been renamed to `writeToAutoComplete()`. ([#140](https://github.com/SchibstedSpain/Barista/issues/140))
+- `clickCheckBoxItem()` methods have been deleted. Use `clickOn()` instead. ([#140](https://github.com/SchibstedSpain/Barista/issues/140))
+- `ListView` and `RecyclerView` actions have been merged into `ListActions` with these common methods ([#146](https://github.com/SchibstedSpain/Barista/issues/146)):
+  - `clickListItem(listId, position)`
+  - `scrollListToPosition(listId, position)`
+  - `clickListItemChild(listId, position, childId)`
+
+## New features
+- `ListView` and `RecyclerView` methods are the same. No need to use different method depending on the implementation. ([#146](https://github.com/SchibstedSpain/Barista/pull/146))
+- `ListView` or `RecyclerView` id is optional when only one instance of `ListView` or `RecyclerView` is displayed in the hierarchy. ([#146](https://github.com/SchibstedSpain/Barista/pull/146)
```
alorma: Missing last parenthesis

screen shot 2017-11-15 at 12 21 36

MIGRATION-2.0.md
```diff
@@ -0,0 +1,43 @@
...
+- Added `assertFocused()` and `assertNotFocused()` methods.
+([#157](https://github.com/SchibstedSpain/Barista/pull/157))
+- Added `assertDisplayed(id, text)` to check some text in a specific view. ([#150](https://github.com/SchibstedSpain/Barista/pull/150))
+- Added `FlakyTestRule` which enables the functionalities from `FlakyActivityTestRule` without forcing the inheritance dependency. ([#160](https://github.com/SchibstedSpain/Barista/pull/160))
+
+## Behaviour changes
+
+#### Error handling
+- Espresso's FailureHandler is not invoked anymore while Barista attempts to do its magic. It will only receive errors that actually make your test fail. This means you can use a custom FailureHandler for things like taking screenshots upon failures. ([#134](https://github.com/SchibstedSpain/Barista/pull/134))
+- We also improved some error messages when things go wrong, so you don't need to decipher what they mean. If you find any message that doesn't make much sense, open en issue!
```
alorma: `open an issue`

MIGRATION-2.0.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Changed methods
+- `click()` methods have been renamed to `clickOn()`. ([#100](https://github.com/SchibstedSpain/Barista/issues/100))
+- `writeToEditText()` method has been renamed to `writeTo()`. ([#140](https://github.com/SchibstedSpain/Barista/issues/140))
+- `writeToAutoCompleteTextView()` method has been renamed to `writeToAutoComplete()`. ([#140](https://github.com/SchibstedSpain/Barista/issues/140))
+- `clickCheckBoxItem()` methods have been deleted. Use `clickOn()` instead. ([#140](https://github.com/SchibstedSpain/Barista/issues/140))
+- `ListView` and `RecyclerView` actions have been merged into `ListActions` with these common methods ([#146](https://github.com/SchibstedSpain/Barista/issues/146)):
+  - `clickListItem(listId, position)`
+  - `scrollListToPosition(listId, position)`
+  - `clickListItemChild(listId, position, childId)`
```
EsteveAguilera: Maybe move one line up the `clickListItemChild` so the two clickList stay together?

README.md
```diff
@@ -92,6 +107,9 @@ assertNotDisplayed("Hello world");
 assertNotDisplayed(R.string.hello_world);
 assertNotDisplayed(R.id.button);
 
+// Is this text displayed in this exact view?
+assertDisplayed(R.id.view, "Hello world")
```
EsteveAguilera: Maybe move it next to the other `assertDisplyed`?

README.md
```diff
@@ -112,14 +130,24 @@ assertChecked("Checked checkbox");
...
 assertUnchecked(R.string.unchecked_checkbox);
 assertUnchecked(R.id.unchecked_checkbox);
 
+// Does this view have the focus?
+assertFocused(R.id.focused_view)
+assertFocused("Button")
+
+// ... or not?
+assertFocused(R.id.focused_view)
+assertFocused("Button")
```
EsteveAguilera: You forgot the **Not** 
`assertNotFocused`

README.md
```diff
@@ -135,25 +163,34 @@ assertThatBackButtonClosesTheApp();
...
 
 The new Marshmallow permissions system requires checking for permissions at runtime. As Espresso can't interact with the system dialog, Barista offers a way to allow permissions when needed.
 
 ```java
 PermissionGranter.allowPermissionsIfNeeded(Manifest.permission.GET_ACCOUNTS);
 ```
 
-## Resetting the app's data before running each test
+## Useful test rules
+Barista includes a set of useful rules test rules to help your tests.
```
EsteveAguilera: Too much "rules" on this line 🤓

Sloy:
Thanks a lot guys. I'm so bad at this 😂  🔪

README.md
rocboronat: rules

README.md
Sloy: I don't think BaristaRule is customizable enough yet :(
For example we need to control excludes from the database rule, or add more custom rules (IdlingResources, memory cache clearing...) and they need to be inserted in a specific part of the RuleChain.

So... yes, I agree that the right approach is to make BaristaRule satisfy everyone's needs, but it's not ready. We need to improve it before.

README.md
Sloy: Other files, like javadocs? 😏  🤡 

I can agree we have too much code in the Readme, we don't need to showcase the whole API...
Is this a blocking thing? Or can we improve it in the future?

README.md
Sloy: Hmmm that kinda breaks the `// blahblahblah | // ... or not` flow.
But my guts also tell me that you're right.

I'll do it anyways 🙈

EsteveAguilera: I know what you mean....

Maybe adding an `assertNotDisplayed(R.id.view, "Hello world")`? But it might not be really useful...

On the other hand, it's kinda strange to have two `assertDisplayed`separated, if they share the method name, they must do "the same"... Are we facing a naming problem (again)?

I don't know 💃

README.md
rocboronat: Then, you're answering your own question 👅 You should explain how to use it with a chain.

README.md
rocboronat: For me, step by step. Finish this PR as soon as possible because it will improve the readme, and then, move to the next thing. Probably we should talk with new Barista users about what they think about the readme before doing a big change on it.

README.md
rocboronat: A savage issue appeared! https://github.com/SchibstedSpain/Barista/issues/164

EsteveAguilera: Leave it to me!

gotta_catch_ em_all

EsteveAguilera: Catch! #165

README.md
```diff
@@ -61,26 +75,27 @@ clickDialogNeutralButton();
...
 swipeViewPagerBack();
-swipeViewPagerBack(R.id.pager);
 
 // Interact with the navigation drawer
-openDrawer(R.id.drawer);
-closeDrawer(R.id.drawer);
+openDrawer();
+closeDrawer();
+openDrawerWithGravity(Gravity.RIGHT);
+closeDrawerWithGravity(Gravity.RIGHT);
```
EsteveAguilera: Another MIMIMI, but what about putting together the two `openDrawer` and `closeDrawer`?

README.md
```diff
@@ -25,29 +46,22 @@ longClickOn(R.id.button);
...
-clickListViewItem(R.id.listview, 4, 5, 6);
-clickRecyclerViewItem(R.id.recycler, 2);
-clickRecyclerViewItem(R.id.recycler, 2, 3, 4);
-clickRecyclerViewItemChild(R.id.recycler, 3, R.id.button);
-clickRecyclerViewItemChild(R.id.recycler, 3, "Button");
+// Operate on ListViews and RecyclerViews indistinctly by position
+clickListItem(R.id.list, 4);
+scrollListToPosition(R.id.list, 4);
+clickListItemChild(R.id.list, 3, R.id.row_button);
+clickListItemChild(R.id.list, 3, "Button");
```
EsteveAguilera: Here comes a MIMIMI... put the `clickListItem...` together

Sloy:
All changes done!
We ready?

PR #167, opened 2017-11-21
Add the scrollTo(R.string.text) method pointing to Barista 1

rocboronat:
Let's fix #22

PR #169, opened 2017-11-21
Add the scrollTo(R.string.text) method pointing to Barista 2

rocboronat:
Same than #167 but pointing to the Barista 2 branch


library/src/main/java/com/schibsted/spain/barista/interaction/BaristaScrollInteractions.kt
```diff
@@ -22,8 +21,8 @@ object BaristaScrollInteractions {
     private val MAX_SCROLL_ATTEMPTS = 50
 
     @JvmStatic
-    fun scrollTo(@IdRes id: Int) {
-        scrollWithMultipleAttempts(withId(id), failAtEnd = true)
+    fun scrollTo(id: Int) {
```
Sloy: Maybe rename to "resId" for consistency with other methods (click), since now we're not only receiving an id.

sample/src/androidTest/java/com/schibsted/spain/barista/sample/ScrollsTest.java
```diff
@@ -23,36 +23,58 @@
...
+    scrollTo(R.string.really_far_away_button);
     assertBottomVisible();
   }
 
   @Test(expected = BaristaException.class)
   public void scrollsOutsideScrollableFails() throws Exception {
     assertTopVisible();
-
     scrollTo(R.id.centered_button);
+    assertTopVisible();
```
Sloy: This is not really wanted. Because `scrollTo` is the method that should fail.

If by any chance the `scrollTo` doesn't fail (like safelyScrollTo) BUT `assertTopVisible` throws a BaristaException, the test will be green while the behavior is broken.

sample/src/androidTest/java/com/schibsted/spain/barista/sample/ScrollsTest.java
```diff
@@ -23,36 +23,58 @@
   public FailureHandlerValidatorRule handlerValidator = new FailureHandlerValidatorRule();
 
   @Test
-  public void scrollsInsideScrollable() {
+  public void scrollsInsideScrollable_byId() {
     assertTopVisible();
-
     scrollTo(R.id.really_far_away_button);
+    assertBottomVisible();
```
Sloy: The blank lines are there to separate the given/when/then parts of the tests. I think they help distinguish that the important method is is `scrollTo`

sample/src/androidTest/java/com/schibsted/spain/barista/sample/ScrollsTest.java
rocboronat: Wow, totally missed the expected when doing the merge :·/ Thanks for spotting!

sample/src/androidTest/java/com/schibsted/spain/barista/sample/ScrollsTest.java
rocboronat: Sorry, I thought in the rule we follow when we worked together, about avoiding to add blank lines when a test just has one line for given, another one for the when and one more for the then.

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaScrollInteractions.kt
rocboronat: On the way!

rocboronat:
Done! 💪

sample/src/androidTest/java/com/schibsted/spain/barista/sample/ScrollsTest.java
Sloy: Um your memory is better than mine, so you're probably right n_n' 
Still, I feel like in this case I read it better with blank lines.

Whatever, thanks for changing it!

rocboronat:
Done at #176 !

PR #171, opened 2017-11-21
Split assertion classes

Sloy:
#The assertions class is growing with lots of unrelated methods.

Since we're breaking the API with 2.0, it's the best time to split them into topic-related classes, just like we do with Interaction methods.

These are the new classes:
screen shot 2017-11-21 at 16 26 03
Some are named after the the concept they assert, some after the only View type they support. Just like the Interactions.

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaAssertions.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaAssertions.kt
@@ -19,26 +19,6 @@ import org.junit.Assert.fail
 object BaristaAssertions {
 
     @JvmStatic
-    fun assertChecked(resId: Int) {
-        onView(resId.resourceMatcher()).check(matches(isChecked()))
-    }
-
-    @JvmStatic
-    fun assertChecked(text: String) {
-        onView(withText(text)).check(matches(isChecked()))
-    }
-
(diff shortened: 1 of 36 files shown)


rocboronat:
@Sloy please, do it after merging the Barista 2 branch 😢. For sure this PR will add value, but you can't refactor and refactor forever and ever. This is like art: it doesn't have an end.

Sloy:
@rocboronat this **must** be done before. It breaks the public API, and we won't have a chance to do that again until 3.0. It's a now or never.

rocboronat:
An idea to remove some merging huge branch: why not just merging the Barista 2 on master, and avoid releasing versions? Or launching versions with the _pre-release_ label.

rocboronat:
We're wasting time and effort (and peace) for the trouble of maintaining two branches at the same time.

Sloy:
We did that to allow releasing 1.x in the meantime (like 1.9).
Although blocking releases until 2.0 is not a bad approach in my opinion (also in favor of my mental health).

rocboronat:
For me, move on. In that way everyone will help you growing up and testing the version 2.0.

Sloy:
Done. Now... 👍  or 👎 ? xd

rocboronat:
Gogo! 👍

PR #175, opened 2017-11-24
Keep the readme style to introduce API methods

rocboronat:
Let's avoid hurting readers eyes 👁‍🗨 👁


rocboronat:
@Sloy remember: it's the lubristil time.

PR #176, opened 2017-11-24
Add the scrollTo(R.string.text) method

rocboronat:
We don't know what happened with #169, so... let's repeat it.


rocboronat:
No prob bro!

PR #177, opened 2017-11-24
Update readme

EsteveAguilera:
I think we can ignore the minor feature version number in the readme

One day after the release of the 2.0 version, we're already on the 2.1, so keeping the `migrate to 2.0` is now obsolete and might be confusing


rocboronat:
Nice for me :·) Let's wait for @Sloy , that knows this kind of fashions. He's a software fashion victim.

Issue #154, opened 2017-11-14
Choose a codestyle


rocboronat:
When we started the project, I think we were using the Square Code Style (https://github.com/square/java-code-styles). But now, the groovy and xml files are 4-4-8. And I'm getting mad.

Let's choose one code style, and reformat the whole project with it. In the meantime, sorry if I restyle the `AndroidManifest.xml`... 😢

Sloy:
We're using Schibsted's internal codestyle

rocboronat:
@sloy what do you think about using a open codestyle? It will help other contributors to contribute. Yesterday I was fixing autoformatter by hand 😵

Sloy:
I'll raise the issue internally to the Schibsted team. I agree we really should either publish our codestyle or use a public one, it's not helpful at all for contributors to use a private codestyle.

EsteveAguilera:
Barista2 is out! It might be a good time to talk about the codestyle :D

Sloy:
How about just using this new shiny official style?: https://android.github.io/kotlin-guides/style.html

EsteveAguilera:
what about this one:
https://github.com/aosp-mirror/platform_development/blob/master/ide/intellij/codestyles/AndroidStyle.xml

EsteveAguilera:
I've just seen that the Square Code Style (https://github.com/square/java-code-styles) has recently included the Android Kotlin style guide updates... so it might be a good choice...

Sloy:
I'll soon include the Schibsted's codestyle into the repo for Kotlin, Java and xml. I don't forget about you @EsteveAguilera ❤️

PR #180, opened 2017-12-01
BaristaSpinnerInteractions converted to kotlin

jamesbluecrow:


jamesbluecrow:
Yes I did. In any case, doesn't Travis do it too?

rocboronat:
Nope @jamesbluecrow, Travis doesn't pass the espresso tests. They are flaky on the emulator that runs on Travis, we don't know why. We're open to PRs to fix it, too 😅

jamesbluecrow:
hahaha. I don't have experience with Travis yet, but if I find time I'll take a look. 
Have we considered on using Firebase test lab? At Trov we use bitrise and connecting it to Test lab is very easy and it works very well.

Sloy:
We tried it some time ago, but didn't work too well either: https://github.com/SchibstedSpain/Barista/pull/5

Maybe we can take another look into it, see if we missed something or it has improved.

jamesbluecrow:
Yes I did run them locally too.
Btw, @Sloy @rocboronat I see that all the test are written in java, is it ok if we convert them to kotlin, or at least the new ones we write them in kotlin?

rocboronat:
IMHO, Kotlin is not really needed or useful in the tests, but I'm OK with both Java or Kotlin ones :·)

jamesbluecrow:
cool, I asked just because it seems that I'm using the same formatter for kotlin but not for java so it is annoying to format the code manually.

rocboronat:
Totally :·/ We opened an issue to fix it: #154

By the way, the most similar codestyle right now is the Square one: https://github.com/square/java-code-styles

Sloy:
@jamesbluecrow right now I see one motive to keep the tests in java: they're or way of making sure the public api is 100% java compatible. (Static methods, overloaded method parameters, etc). Although I'm sure at some point we will have to migrate them, I'm not sure if we're there yet. 

It's a personal opinion though, we haven't really discussed it at all yet.

rocboronat:
Oh, it's a good reason, never thought about it.

jamesbluecrow:
I didn't think about that and it makes a lot of sense, smart thinking @Sloy 👍

Issue #181, opened 2017-12-02
Missing interactions for SeekBar


jamesbluecrow:
The title is self explanatory, there are no functions to interact with a `SeekBar`. I think it could be nice to have at least:
- BaristaSeekBarInteractions.swipeRight()
- BaristaSeekBarInteractions.swipeLeft()

jamesbluecrow:
I've created a PR to add them https://github.com/SchibstedSpain/Barista/pull/182

PR #170, opened 2017-11-21
Add assertContains matcher

alorma:
As reported by @rocboronat  on #166 

I've added `assertContains` method, to check if a textview contains text, in any part of the `getText()`

PR #182, opened 2017-12-02
Adding interactions for SeekBar views

jamesbluecrow:
Adding interactions for SeekBar views: https://github.com/SchibstedSpain/Barista/issues/181


rocboronat:
@jamesbluecrow Hi! Thanks for your contribution! By the way, it seems that the PR doesn't pass the checks. Can you check them?

Thanks!

jamesbluecrow:
@rocboronat do you know when the codestyle issue https://github.com/SchibstedSpain/Barista/issues/154  will be resolved? Travis is complaining for the checkstyle :(

I will fix it in a moment :)

jamesbluecrow:
I like your idea @alorma. Will include those changes in this PR

alorma:
lovely new interactions! let's wait for @sloy feedback!

sample/src/androidTest/java/com/schibsted/spain/barista/sample/SeekBarAssertionsTest.java
```diff
@@ -0,0 +1,34 @@
...
+    assertProgress(R.id.seek_bar_progress_27, 27);
+  }
+
+  @Test
+  public void checkMinProgress() {
+    assertProgressIsMin(R.id.seek_bar_min_progress);
+  }
+
+  @Test
+  public void checkMaxProgress() {
```
rocboronat: Can you also add the checks that the assertion fails if the data is not the expected one? Something like I did on this commit of another PR: https://github.com/SchibstedSpain/Barista/pull/170/commits/fc159d8798df422ab88a7834410f8db6193f37fb

library/src/main/java/com/schibsted/spain/barista/internal/viewaction/SeekBarActions.kt
```diff
@@ -0,0 +1,48 @@
...
+  fun setProgressToMin(): ViewAction = setProgress(0)
+
+  fun setProgressToMax(): ViewAction {
+    return object : ViewAction {
+      override fun getConstraints(): Matcher<View> {
+        return allOf(isDisplayed(), isAssignableFrom(SeekBar::class.java))
+      }
+
+      override fun getDescription(): String {
+        return "Set SeekBar progress."
```
rocboronat: Not a huge issue for me, but we can extract this text to a constant called `DESCRIPTION`

rocboronat:
And another thing: remember to add those actions and assertions to the readme 🙌

sample/src/androidTest/java/com/schibsted/spain/barista/sample/SeekBarAssertionsTest.java
jamesbluecrow: It makes sense 👍

library/src/main/java/com/schibsted/spain/barista/internal/viewaction/SeekBarActions.kt
jamesbluecrow: Sounds good, I'll extract it 👍

library/src/main/java/com/schibsted/spain/barista/internal/matcher/SeekBarAssertions.kt
```diff
@@ -0,0 +1,60 @@
...
+import android.view.View
+import android.widget.SeekBar
+import com.schibsted.spain.barista.internal.failurehandler.SpyFailureHandler
+import com.schibsted.spain.barista.internal.failurehandler.description
+import org.hamcrest.Matcher
+import org.hamcrest.MatcherAssert.assertThat
+import org.hamcrest.Matchers
+import org.hamcrest.Matchers.`is`
+
+object SeekBarAssertions {
```
Sloy: Random thought: 
Progress is actually not part of `SeekBar` but `ProgressBar` which is the parent class. Does it make sense that the assertions are also compatible with `ProgressBar`? What do you think?

The interaction is similar, but in that case since the user can't change the progress via the UI, I think it's better to limit it to the SeekBar.

sample/src/androidTest/java/com/schibsted/spain/barista/sample/SeekBarAssertionsTest.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Test
+    public void checkMinProgress() {
+        assertProgressIsMin(R.id.seek_bar_min_progress);
+    }
+
+    @Test(expected = AssertionFailedError.class)
+    public void checkMinProgress_failsWhenProgressNotMin() {
+        assertProgress(R.id.seek_bar_min_progress, 10);
```
rocboronat: This should be `assertProgressIsMin(R.id.seek_bar_progress_27);`

sample/src/androidTest/java/com/schibsted/spain/barista/sample/SeekBarAssertionsTest.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Test
+    public void checkMaxProgress() {
+        assertProgressIsMax(R.id.seek_bar_max_progress);
+    }
+
+    @Test(expected = AssertionFailedError.class)
+    public void checkMinProgress_failsWhenProgressNotMax() {
+        assertProgress(R.id.seek_bar_max_progress, 10);
```
rocboronat: And this should be `assertProgressIsMax(R.id.seek_bar_progress_27);`. By the way, the name of the test should be Max and not Min :)

sample/src/androidTest/java/com/schibsted/spain/barista/sample/SeekBarAssertionsTest.java
jamesbluecrow: You are completely right, this is what happens when someone tries to rush 😕

library/src/main/java/com/schibsted/spain/barista/internal/matcher/SeekBarAssertions.kt
```diff
@@ -0,0 +1,60 @@
...
+  @JvmStatic
+  fun assertProgress(expectedProgress: Int): ViewAssertion {
```
Sloy: Have you consider implementing this with a BoundedMatcher instead of a ViewAssertion? 
Matchers have a higher-level api, and automatically handle thing like casting the view, throwing exceptions and creating nice error messages.

You can create them anonymously, or with their own class like for example:
```kotlin
class ProgressMatcher(private val expectedProgress: Int): BoundedMatcher<View, ProgressBar>(ProgressBar::class.java) {

    override fun matchesSafely(progressBar: ProgressBar) = progressBar.progress == expectedProgress

    override fun describeTo(description: Description) {
      description.appendText("with progress $expectedProgress")
    }

}
```

The difference in usage is that with a ViewAssertion you do:
```
onView(...).check(myAssertion)
```
but with a matcher you do the more common:
```
onView(...).check(matches(myMatcher))
```

library/src/main/java/com/schibsted/spain/barista/internal/matcher/SeekBarAssertions.kt
```diff
@@ -0,0 +1,60 @@
...
+    return ViewAssertion { view, noViewFoundException ->
+      if (noViewFoundException != null) {
+        throw noViewFoundException
+      }
+      val seekBar = view as SeekBar
+      assertThat(seekBar.progress, `is`(seekBar.max))
+    }
+  }
+
+  private fun assertDisplayed(matcher: Matcher<View>) {
```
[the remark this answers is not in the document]
jamesbluecrow: Whoa! Not sure why the heck this code is here. I'll remove it!

library/src/main/java/com/schibsted/spain/barista/internal/matcher/SeekBarAssertions.kt
jamesbluecrow: I didn't think about it, but I do agree that there is no point on having assertions for a view whose progress the dev has no control. I guess that if int the future someone finds a case where it is needed adding support shouldn't be complicated.

library/src/main/java/com/schibsted/spain/barista/internal/matcher/SeekBarAssertions.kt
jamesbluecrow: I will take a look, I based the current implementation in `RecyclerViewItemCountAssertion` where it inherits from `ViewAssertion` but `BoundedMatcher` seems to be cleaner.

Sloy: Ouch you're right, I forgot we had that one as a ViewAssertion. It's from the days we didn't know that much about Espresso, sorry :P
I will try and migrate to a matcher for consistency.

jamesbluecrow:
@Sloy @rocboronat I've addressed all the comments, let me know if you have more suggestions

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaSeekBarViewInteractions.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.schibsted.spain.barista.interaction.BaristaScrollInteractions.safelyScrollTo
+import com.schibsted.spain.barista.internal.viewaction.SeekBarActions.setProgress
+import com.schibsted.spain.barista.internal.viewaction.SeekBarActions.setProgressToMax
+import com.schibsted.spain.barista.internal.viewaction.SeekBarActions.setProgressToMin
+
+object BaristaSeekBarInteractions {
+  @JvmStatic
+  fun setProgress(@IdRes seekBarId: Int, progress: Int) {
+    safelyScrollTo(seekBarId)
+    Espresso.onView(ViewMatchers.withId(seekBarId)).perform(setProgress(progress))
```
Sloy: I'm thinking... both the BaristaInteraction and the view Action static methods have the same name. Even though one it's an internal method, to the outside is public.
We've had problems with this before (https://github.com/SchibstedSpain/Barista/issues/100).

I'm not sure which method could we change to what. Currently they both make sense :(
Opinions? @jamesbluecrow @rocboronat @alorma @SmasSive?

library/src/main/java/com/schibsted/spain/barista/assertion/BaristaSeekBarAssertions.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import android.support.annotation.IdRes
+import android.support.test.espresso.Espresso.onView
+import android.support.test.espresso.assertion.ViewAssertions.matches
+import com.schibsted.spain.barista.internal.matcher.DisplayedMatchers.displayedWithId
+import com.schibsted.spain.barista.internal.matcher.ProgressBarMatchers.assertProgress
+import com.schibsted.spain.barista.internal.matcher.ProgressBarMatchers.assertProgressIsMax
+import com.schibsted.spain.barista.internal.matcher.ProgressBarMatchers.assertProgressIsMin
+
+object BaristaSeekBarAssertions {
```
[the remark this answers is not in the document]
jamesbluecrow: I thought about it but I wasn't sure if it was going to be intuitive.
I only find this interactions/assertions useful for `SeekBars` and maybe people are not going to know they are available if it says `ProgressBar`.
 In fact until you told me that `SeekBar` was a `ProgressBar` child I didn't know it was hehe.
Although it can be mentioned in README file. What do you think? Should we go with `ProgressBar` in the name and mention that it supports `ProgressBar` and `SeekBar` in the README?

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaSeekBarViewInteractions.kt
jamesbluecrow: That is a good point, I'm thinking of different alternatives like:

```
- setProgress() vs changeProgressTo() vs swipeProgressTo()
- setProgressToMax() vs changeProgressToMax() vs swipeProgressToMax()
- setProgressToMin() vs changeProgressToMin() vs vs swipeProgressToMin()
```
for some reason I think that I like more the `swipeProgressTo()` pattern for `BaristaSeekBarInteractions` since it represents what it is actually doing, swiping the progress.

library/src/main/java/com/schibsted/spain/barista/internal/matcher/ProgressBarMatchers.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    override fun matchesSafely(progressBar: ProgressBar) = progressBar.progress == expectedProgress
+
+    override fun describeTo(description: Description) {
+      description.appendText("should be equal to progress $expectedProgress")
+    }
+  }
+
+  @JvmStatic
+  fun assertProgress(expectedProgress: Int) = SeekBarProgressMatcher(expectedProgress)
```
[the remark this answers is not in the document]
jamesbluecrow: Makes total sense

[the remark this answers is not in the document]
jamesbluecrow: I'm modifying it to be: `withProgress()`, `withMinProgress()` and `withMaxProgress()`

library/src/main/java/com/schibsted/spain/barista/internal/matcher/ProgressBarMatchers.kt
```diff
@@ -0,0 +1,38 @@
...
+
+  @JvmStatic
+  fun assertProgressIsMin() = SeekBarProgressMatcher(0)
+
+  @JvmStatic
+  fun assertProgressIsMax(): BoundedMatcher<View, SeekBar> {
+    return object : BoundedMatcher<View, SeekBar>(SeekBar::class.java) {
```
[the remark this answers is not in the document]
jamesbluecrow: True 👍

library/src/main/java/com/schibsted/spain/barista/assertion/BaristaSeekBarAssertions.kt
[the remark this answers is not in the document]
SmasSive: I'd go for `setProgress` for the internal method because it sounds more techie (setter and getter) and `changeProgress` for the public one because it implies not only setting the progress, maybe in a future implies more actions. What do you think?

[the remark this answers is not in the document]
Sloy: @SmasSive I think you commented on the wrong thread :P 
You're probably talking about https://github.com/SchibstedSpain/Barista/pull/182#discussion_r156118773

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaSeekBarViewInteractions.kt
Sloy: From @SmasSive comment:
> I'd go for setProgress for the internal method because it sounds more techie (setter and getter) and changeProgress for the public one because it implies not only setting the progress, maybe in a future implies more actions. What do you think?

I'm going to use your argument against you :P 
Since `setProgressTo` is more techie it helps discoverability. In the end Barista users are developers. This name is straightforward, short and simple. No confusion there. Maybe `setProgressTo` fits also.

I have trouble with `swipeProgressTo`. For some reason I see swiping as a less precise action, not the kind of think you do with a SeekBar. I don't see it as much intuitive. BUT that may just be my English level anyways.

To write down a formal opinion, I would go for
- `setProgressTo` for the public methods, because of simplicity
- `setSeekBarProgressTo` for the internal ViewActions, because it exposes implementation a bit.


All these arguments are very thin though. Aw man, why did naming have to be so hard u_u

library/src/main/java/com/schibsted/spain/barista/assertion/BaristaSeekBarAssertions.kt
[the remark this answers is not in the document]
Sloy: In the case of assertions @jamesbluecrow I think they also make sense for ProgressBar, not just SeekBar. 
I can think of a couple scenarios:
- You have a long wizard where a ProgressBar shows your current page and you want to test that it updates.
- Something like Linkedin profile completion progress, you want to test it shows the right percentage. 

Just some examples to say that in my opinion progress assertion is useful for both SeekBar and ProgressBar. Because of that I would just call them ProgressAssertions, which talks about the concept but not the concrete View.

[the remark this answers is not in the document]
SmasSive: Yes buddy, some mobile issues... Don't try to answer a huge commented PR like this by phone xD

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaSeekBarViewInteractions.kt
SmasSive: I agree with you here @Sloy , what I was tending to explain is that `setProgress` is very concise, the name says it does just one thing (or should it do) and `swipeProgressTo` something more open that could imply more things, so gathering all the info and opinions... I think in your last comment you nailed it! 🗡

rocboronat:
@jamesbluecrow sorry for my late review! I was moving to another home 😅

library/src/main/java/com/schibsted/spain/barista/assertion/BaristaSeekBarAssertions.kt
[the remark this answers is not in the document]
jamesbluecrow: Fair enough @Sloy , I think those are valid scenarios. I will update the name today

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaSeekBarViewInteractions.kt
jamesbluecrow: SGTM too 👍 
I will update it today!

jamesbluecrow:
Not a problem at all @rocboronat , as you can see we've been busy talking on this PR :)
Where did you move from/to?

jamesbluecrow:
I did run Tests & Checks so it should be good to go now.

Issue #183, opened 2017-12-02
Problem when trying to click a button that is behind the keyboard


jamesbluecrow:
In a test for our login screen, I found an "error" in the library when writing into an `EditText/AutoCompleteTextView`, the "problem" appears when the screen has a button at the bottom and it is hidden behind the keyboard.

![screenshot_1512221877](https://user-images.githubusercontent.com/8219006/33516246-7321ec74-d767-11e7-918f-7dd9e2117cc9.png)
![screenshot_1512221887](https://user-images.githubusercontent.com/8219006/33516247-7334a99a-d767-11e7-9e14-b011ea9f4f6c.png)

We have a test like this:

```
BaristaEditTextInteractions.writeTo(R.id.email, "barista@gmail.com")
BaristaEditTextInteractions.writeTo(R.id.password, "")
BaristaClickInteractions.clickOn(R.id.submit)
```

But we get this is the error because we are trying to click a button that is not visible and it's hidden behind the keyboard.

```
com.schibsted.spain.barista.internal.failurehandler.BaristaException: 
Could not click on view with id: com.trov.maui.staging:id/submit
```

My proposal would be to close by default the keyboard after the text has been written, also it could provide the option to submit from the keyboard.

I've already created some extensions for this so I will open a PR to show the proposed solution.'

rocboronat:
Hey, good issue! By the way, don't you think that you have an issue in your app? I mean, what if you just put the login button below the password, instead of doing it at the bottom of the screen? In that way, you will save a tap to the final customer.

Oh, and send that proposal, for sure! :·D

jamesbluecrow:
hahaha, that is something that we would need to tell to the design team ;) Thanks for the suggestion!

Another option is to support `Submit` from the keyboard (which right now I'm not sure we do).

Here is the PR: https://github.com/SchibstedSpain/Barista/pull/184

rocboronat:
Hey, what do you think about just adding the methods to close the keyboard to the Barista API? In that way, you can close the keyboard, but usually in the same way than you will find a bug in his app thanks to the Espresso test. I mean, your proposal will give a green in a test, but actually what you are doing is bypassing a bug.

Let me think about it after the weekend! :·)

jamesbluecrow:
I'm ok with adding support to close the keyboard as part of the Barista API although I believe that here we are talking about two different things.
The fact that a button is behind the keyboard is not a bug since we can decide how we want the app to resize the screen.
[...]

PR #184, opened 2017-12-02
Add keyboard actions to BaristaEditTextInteractions and BaristaAutoCompleteTextViewInteractions

jamesbluecrow:
This PR tries to solve https://github.com/SchibstedSpain/Barista/issues/183

It seems to me that by default it should close the keyboard after writing.


rocboronat:
To other reviewers: don't hesitate to read the #183 conversation

EsteveAguilera:
I think it's not a good idea to overwrite the current interactions... you'll break a lot of tests!

Why not add new interactions with the KeyboardActions instead?

PR #185, opened 2017-12-04
Add keyboard interactions

jamesbluecrow:
I've created a new class named `BaristaKeyboardInteractions` that has two functions:
- close()
- pressImeActionButton()

It is possible to close the keyboard without providing a viewId so that's cool, although there is no way to test that the keyboard has been closed without adding some "hacky" code that I'm not sure we want. Let me know what you guys think.


Sloy:
Let's move the implementation discussion here!

> there is no way to test that the keyboard has been closed without adding some "hacky" code

Actually, could we reproduce a scenario similar to the one you reported in your issue? Some view that is hidden by the keyboard, and when you close the keyboard becomes visible.

Sloy:
For reference, this PR is related to https://github.com/SchibstedSpain/Barista/issues/183

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaKeyboardInteractions.kt
```diff
@@ -0,0 +1,24 @@
...
+
+object BaristaKeyboardInteractions {
+
+    @JvmStatic
+    fun close() {
+        onView(isRoot()).perform(closeSoftKeyboard())
+    }
+
+    @JvmStatic
+    fun pressImeActionButton(@IdRes editTextId: Int) {
```
Sloy: An idea, if you want the challenge:
we could try to press on the currently focused EditText, avoiding the need to specify a view id. It would be similar to what we do with `findDrawerMatcher()` in `BaristaDrawerAssertions`, for example, using a `hasFocus() + isAssignableFrom(EditText)` matcher.

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaKeyboardInteractions.kt
```diff
@@ -0,0 +1,24 @@
...
+import android.support.test.espresso.action.ViewActions.pressImeActionButton
+import android.support.test.espresso.matcher.ViewMatchers.isRoot
+import android.support.test.espresso.matcher.ViewMatchers.withId
+import com.schibsted.spain.barista.interaction.BaristaScrollInteractions.safelyScrollTo
+import com.schibsted.spain.barista.internal.matcher.DisplayedMatchers.displayedAnd
+
+object BaristaKeyboardInteractions {
+
+    @JvmStatic
+    fun close() {
```
Sloy: Suggestion: Thinking about api usage, if people want to statically import Barista methods to avoid the long class prefix, this should be called `closeKeyboard()` to better predict what is it about.

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaKeyboardInteractions.kt
jamesbluecrow: I will check that code out and modify the function accordingly 👍

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaKeyboardInteractions.kt
jamesbluecrow: 👍

jamesbluecrow:
I thought about it https://github.com/SchibstedSpain/Barista/pull/185#issuecomment-349031432 but that would not be testing that the keyboard is hidden by that method, in fact for devices that don't use virtual keyboard the test would pass but it would not mean that the keyboard was hidden because it would pass without calling the `closeKeyboard()` function.

I don't mind adding a test like that if that is what we want though

jamesbluecrow:
@Sloy I've accepted the challenge of adding support for `pressImeActionButton()` for the focused view without providing an `id`, but in order to make it work correctly if the action needed a scroll I've been forced to duplicate some code already existent on `BaristaClickInteractions.performClickTypeOnMatcher(). I do believe that code should/could be extracted to a place where it can be reused but in my opinion that would be out of the scope of this PR.

Sloy:
I see what you mean about the duplication. You're right, that should be extracted to some generic construct and probably used on a lot more Interactions. But I also agree that's not the purpose of this PR :) Nice work!

Sloy:
I've been thinking about your comment on how to test closeKeyboard() and how else we could do it. 
I was trying to implement something to check the keyboard status from the `InputMethodManager`. But reading the code from Espresso's `CloseKeyboardAction`, I see the action already verifies the status after performing.

I can see arguments for and against doing the test case that I suggested. Since is a shortcut for `CloseKeyboardAction` and it already verifies the keyboard is closed, my opinion goes to: let's leave the test out, and if we ever find a bug here we'll cover the method then.

jamesbluecrow:
My pleasure, you guys did the big effort to have the idea and create this amazing library 😄

jamesbluecrow:
I just realized that I never added the new methods to the `README`. Now they are there :)

Sloy:
Wops, good catch!

PR #192, opened 2017-12-15
Fixes wrong error handling on assertDisplayed

Sloy:
While developing on https://github.com/Sloy/android-screenshot-reporter, which uses a FailureHandler to take screenshots on errors, I realised that the magic in assertDisplayed method wasn't behaving as expected.

**How should it work:**
When you call `assertDisplayed(id|text)` and the condition doesn't match -> you should receive **one and only one** error in your global Espresso's FailureHandler. Either the default `DefaultFailureHandler` or your own setted with `Espresso.setFailureHandler(custom)`. 
That happens thanks to `SpyFailureHandler.resendFirstError(...)`.

**How is it working now:**
When you call `assertDisplayed(id|text)` and the condition doesn't match -> The exception is simply thrown and your FailureHandler doesn't get notified.

The cause was a wrong logic on the try/catch flow. We were only rethrowing in case of `AmbiguousViewMatcherException`, which is not often the case.

Well, this PR fixes it.

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaVisibilityAssertions.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaVisibilityAssertions.kt
@@ -36,18 +36,23 @@ object BaristaVisibilityAssertions {
         onView(withId(resId)).check(matches(withText(text)))
     }
 
+    /**
+     * Attempts to find the view with multiple conditions:
+     * 1. Simplest case
+     * 2. More than one view
+     */
     private fun assertDisplayed(matcher: Matcher<View>) {
         val spyFailureHandler = SpyFailureHandler()
         try {
             onView(matcher)
                     .withFailureHandler(spyFailureHandler)
                     .check(matches(isDisplayed()))
-        } catch (multipleViews: AmbiguousViewMatcherException) {
+        } catch (firstError: Exception) {
             try {
                 onView(HelperMatchers.firstViewOf(allOf(matcher, isDisplayed())))
                         .withFailureHandler(spyFailureHandler)
                         .check(matches(isDisplayed()))
-            } catch (notDisplayedError: NoMatchingViewException) {
+            } catch (secondError: Exception) {
                 spyFailureHandler.resendFirstError("View ${matcher.description()} wasn't displayed on the screen")
             }
         }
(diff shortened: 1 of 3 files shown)


Sloy:
I know @SmasSive has some bad comments about that ugly multi-try/catch structure. And I agree, we should refactor and extract it for easier reutilization, and I will. But this PR is just about the bug.

Sloy:
Humm this breaks one test:
```
@Test(expected = NoMatchingViewException.class)
  public void checkTextViewContainsText_withoutViewId_failsWhenNeeded() {
    assertContains("unexisting text");
  }
```
Because it's now throwing a `BaristaException` instead.
That makes me wonder if BaristaException is a good thing or not. Maybe we should just re-throw the original Espresso error without wrapping it. 
But it helps when we need to use a custom error message, like ""No ListView or RecyclerView found in the hierarchy".

library/src/main/java/com/schibsted/spain/barista/assertion/BaristaVisibilityAssertions.kt
```diff
@@ -36,18 +36,23 @@ object BaristaVisibilityAssertions {
         onView(withId(resId)).check(matches(withText(text)))
     }
 
+    /**
```
SmasSive: I hope in the next refactor this comment is not necessary anymore... 🙏

Sloy: It's not because we'll move the code somewhere else. But I was thinking about leaving some similar documentation. What problem do you see with the comment? Don't you think it's a good thing to document the not-so-obvious behaviors of our methods?

ivngrz: Having documentation comments is actually a good thing, IMO, but they aren't really all that useful in a private method. It would make more sense to comment the interface the users will actually work with.

SmasSive: Yes it is but I was talking about destroying this so ugly code and make it beauty, so so beauty that you won't need this documentation 👍

Sloy: Good luck with that 🤣 
Let's talk about it in the refactor PR, maybe we can achieve it.

PR #196, opened 2017-12-15
Added clickOn longClickOn wihtId and withText

jamesbluecrow:

Issue #107, opened 2017-07-17
Issue when writing in an EditText that appears in another page of a ViewPager


rocboronat:
Same than #68 , but writing into `EditText`. That's strange... looked at the source, and it should work. It's EDD time.

Just for the information, a piece of code to showcase the issue. The click works and the write not:

```java
click(R.id.et_postal_code)
writeToEditText(R.id.et_postal_code, "17750")
```

Sloy:
This is supposed to be fixed after https://github.com/SchibstedSpain/Barista/pull/198
If you find anything any other issue reopen it @rocboronat :D

Issue #46, opened 2017-04-19
Same than 


rocboronat:
Same than #37 , but with `assertNotDisplayed()`

Sloy:
I wonder if we should implement a generic solution for this. Sounds like all actions and assertions should have the same behavior when dealing with scroll or multiple views on screen.

Issue #68, opened 2017-05-17
click(id) doesn't work when multiple views in the hierarchy


MarinaTov:
When I click `close_button`:

![1495025046092](https://cloud.githubusercontent.com/assets/28757304/26154905/2dcbfe50-3ac6-11e7-88d0-3ebe666163a4.jpg)

I get error `android.support.test.espresso.AmbiguousViewMatcherException: 'with id: com.schibsted.android.rocket.kufar.debug:id/close_button' matches multiple views in the hierarchy.
Problem views are marked with '****MATCHES****' below.`

![screen shot 2017-05-17 at 5 38 45 am](https://cloud.githubusercontent.com/assets/28757304/26154912/360b6998-3ac6-11e7-9908-b3abf5831516.png)

rocboronat:
Hi @MarinaTov !

It seems that Espresso is trying to express the idea that multiple Views has the ID you're trying to click. So, it doesn't know what View is the View you want to click.

To fix the issue, we have different options there:
- Give different names to that different buttons. It's the simpler solution, and IMHO, something that improves the maintainability of the whole project. But it could be an issue if it's placed inside a RecyclerView, like your case.
- Use `BaristaRecyclerViewActions.clickRecyclerViewItemChild(R.id.user_preference_category_recycler_view, 0, R.id.close_button)`, that will try to tap `R.id.close_button` of the first child inside `R.id.user_preference_category_recycler_view`.
- Update Barista to the last version (1.0.1), as it only clicks on visible items. It could be an easy fix to this scenario. But the issue will appear again when you place two childs inside the recycler with the `R.id.close_button` visible.

Let me keep this question open to track if the answer helps you fixing it! :·)

MarinaTov:
version 1.0.1 resolved problem
Thanks a lot

rocboronat:
You're welcome! :·)

RajaniAchrya:
Hi, I'm getting id/phoneEditText' matches multiple views in the hierarchy. Problem views are marked with '****MATCHES****' below.
where there's only one view with this id.

ShaxGOD:
How can we just interact with id ? like id/phoneEditText[0] - is that possible ?

PR #198, opened 2017-12-18
Introduce generic assertAnyView function

Sloy:
## TL;DR:
This method allows asserting that any view on the screen matches a certain condition, without getting the infamous Espresso error `AmbiguousViewMatcherException`.

## Rationale
We say we do magic for things like views below a scroll or duplicated views on screen. But actually that only work for some Assertions and Interactions, not all of them implement it. E.g.: https://github.com/SchibstedSpain/Barista/issues/107 https://github.com/SchibstedSpain/Barista/issues/46

## Solution
In this PR I propose to extract the assertion code to one place, and reuse it everywhere it's needed. (The same should be done for Interactions, see https://github.com/SchibstedSpain/Barista/pull/196).
I created a function to be used like this:

```kotlin
assertAnyView(withId(R.id.button), isDisplayed())
```

We don't usually expose kotlin APIs, but since this method is intended for internal usage I also added an extension function alias:
```kotlin
withId(R.id.button).assertAny(isDisplayed())
```

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/internal/MagicAssert.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/internal/MagicAssert.kt
@@ -1,0 +1,63 @@
+package com.schibsted.spain.barista.internal
+
+import android.support.test.espresso.Espresso.onView
+import android.support.test.espresso.assertion.ViewAssertions
+import android.support.test.espresso.matcher.ViewMatchers.withText
+import android.view.View
+import com.schibsted.spain.barista.internal.failurehandler.SpyFailureHandler
+import com.schibsted.spain.barista.internal.failurehandler.description
+import com.schibsted.spain.barista.internal.matcher.HelperMatchers.firstViewOf
+import com.schibsted.spain.barista.internal.util.resourceMatcher
+import org.hamcrest.Matcher
+import org.hamcrest.Matchers.allOf
+import org.hamcrest.Matchers.not
+
+/**
+ * Extension function alias for [magicAssertOnView] supporting both Id or String resources.
+ */
+infix fun Int.magicAssert(condition: Matcher<View>) {
+    this.resourceMatcher() magicAssert condition
+}
+
+/**
+ * Extension function alias for [magicAssertOnView] using [withText] as viewMatcher
+ */
+infix fun String.magicAssert(condition: Matcher<View>) {
+    withText(this) magicAssert condition
+}
+
+/**
+ * Extension function alias for [magicAssertOnView]
+ */
+infix fun Matcher<View>.magicAssert(condition: Matcher<View>) {
+    magicAssertOnView(viewMatcher = this, condition = condition)
+}
+
+/**
+ * Performs an assertion of a [condition] on a view described by [viewMatcher].
+ *
+ * Attempts to assert using multiple scenarios for the [viewMatcher]:
+ * 1. Just one view matches the [viewMatcher].
+ * 2. Multiple views match the [viewMatcher]: will pass if at least one of them matches the [condition].
+ */
+fun magicAssertOnView(viewMatcher: Matcher<View>, condition: Matcher<View>) {
+    val spyFailureHandler = SpyFailureHandler()
+    try {
+        onView(viewMatcher)
+                .withFailureHandler(spyFailureHandler)
+                .check(ViewAssertions.matches(condition))
+    } catch (firstError: RuntimeException) {
+        try {
+            onView(firstViewOf(allOf(viewMatcher, condition)))
+                    .withFailureHandler(spyFailureHandler)
+                    .check(ViewAssertions.matches(condition))
+        } catch (secondError: RuntimeException) {
+            spyFailureHandler.resendFirstError("View ${viewMatcher.description()} wasn't displayed on the screen")
+        }
+    }
+}
+
+/**
+ * Allow using `!viewMatcher` as alias for `not(viewMatcher)`
+ */
+operator fun Matcher<View>.not(): Matcher<View> = not(this)
--- a/library/src/main/java/com/schibsted/spain/barista/internal/MagicAssert.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/internal/MagicAssert.kt
@@ -64,8 +64,3 @@ private fun tryToAssert(viewMatcher: Matcher<View>, condition: Matcher<View>, sp
             .withFailureHandler(spyFailureHandler)
             .check(ViewAssertions.matches(condition))
 }
-
-/**
- * Allow using `!viewMatcher` as alias for `not(viewMatcher)`
- */
-operator fun Matcher<View>.not(): Matcher<View> = not(this)
--- a/library/src/main/java/com/schibsted/spain/barista/internal/MagicAssert.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/internal/MagicAssert.kt
@@ -2,7 +2,6 @@ package com.schibsted.spain.barista.internal
 
 import android.support.test.espresso.Espresso.onView
 import android.support.test.espresso.assertion.ViewAssertions
-import android.support.test.espresso.matcher.ViewMatchers.withText
 import android.view.View
 import com.schibsted.spain.barista.internal.failurehandler.SpyFailureHandler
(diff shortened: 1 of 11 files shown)


library/src/main/java/com/schibsted/spain/barista/internal/MagicAssert.kt
```diff
@@ -0,0 +1,63 @@
...
+ * 1. Just one view matches the [viewMatcher].
+ * 2. Multiple views match the [viewMatcher]: will pass if at least one of them matches the [condition].
+ */
+fun magicAssertOnView(viewMatcher: Matcher<View>, condition: Matcher<View>) {
+    val spyFailureHandler = SpyFailureHandler()
+    try {
+        onView(viewMatcher)
+                .withFailureHandler(spyFailureHandler)
+                .check(ViewAssertions.matches(condition))
+    } catch (firstError: RuntimeException) {
```
Sloy: @SmasSive look! A try/catch in the wild!

I tried implementing this with some kind of [Try monad](http://kategory.io/docs/datatypes/try/) to avoid anidating these try/catch, and it worked. But I'm not sure it's worth that complexity just for one method 😕  (or two in the future).

Sloy: Here's what I was able to implement with Kategory.
captura de pantalla 2017-12-18 a las 9 17 20

It might be simplified even more with some custom structure instead of this one. Not sure if that's enough though. 
Thoughts? 💭

SmasSive:
Hi, mimimi here.
I don't agree with infix function, maybe is my lack of knowledge of Kotlin or that I didn't use it enough but for me is not natural to use the API like this `withId(R.id.button) magicAssert not(isDisplayed())` I think is better to use it semantically like this `withId(R.id.button).magicAssert(not(isDisplayed()))` but it's ok since we don't force anyone to use it only in a certain way, remember I am just a mimimi.

On the other hand, I don't agree either with the alias for the `not` operator, users are used to Espresso and they write things like `not`, so I don't see any reason to convert it to `!`, I think it only can confuse users about its usage.

Sloy:
Not mimimi at all. I did those things _"because I could"_. Now it's the moment to think if they make sense or not.
Yep, I too think the `not` operator can add confusion. 
The infix... I like it for some reason. But it's not rational, just a personal feeling. I don't have real arguments for or against it.

How about step 3?
Would you still do the extension `R.id.button.magicAssert(not(isDisplayed()))` for ids?

Sloy:
Actually, what I think I like of the infix is that is clearly separates the view finder and the actual condition. In a long line like `withId(R.id.button).magicAssert(not(isDisplayed()))` with many dots and parenthesis, you have to stop (a bit) and see where those two parts are separated. With the infix is easier to spot. 
But anyways, it's true I haven't seen many APIs using it either, don't know what's the good practice around it.

library/src/main/java/com/schibsted/spain/barista/internal/MagicAssert.kt
SmasSive: I do love the implementation with Kategory, it simplifies a lot the code. I'd love it even more if you separate the `Try`s in methods, just to improve readability, something like:
```kotlin
tryToAssertView()
  .recoverWith { tryToAssertFirstViewOf() }
  .getOrElse { spy... }
```
What do you think?

Sloy: You know I hate you, right? 🤣 
Ok let me see what I can do.

SmasSive:
Step 3 is more like the `not`, I think we should ask ourselves... Do we want to abstract as much as possible the Espresso API? Or do we want that Barista is just a "enabler" of Espresso? The answer to that question will light our path, if we want to create our own semantic, abstract as much as possible the usage of Espresso, let's go for `!` and id's. If we don't, if we want that Barista remains as the sugar of Espresso, we should not add too many changes.

Sloy:
So you go for the more espresso-like syntax.

library/src/main/java/com/schibsted/spain/barista/internal/MagicAssert.kt
Sloy: Here are a couple of examples I've put together.

With Kategory:
captura de pantalla 2017-12-19 a las 10 55 19
With a custom DSL-ish:
captura de pantalla 2017-12-19 a las 12 03 55

I like how the second one looks. But again, I'm not sure if we're being pragmatic here. Seems like a lot of complexity for something that will be use in just two places.

SmasSive: I like the second option too but the key here I think is to see how it is implemented, if it adds a lot of complexity I'd discard it but if it doesn't, go for it!
BTW, I invoke some of the usual contributors like @rocboronat @alorma @sergiandreplace to light us up here :dancing_men:

Sloy: Well... For the implementation I could use a hand.

This is something that works, not too obvious though:
captura de pantalla 2017-12-19 a las 14 45 49

Sloy: I'm starting to think the initial double try/catch is the best solution 🤣

SmasSive: Ok, looks complex... So my call would go for the first implementation with Kategory.
Please please please, don't kill me ❤️

rocboronat:
Let me share an idea: we love magic when interacting with the views, but we must hate magic when asserting. For example, if I assert if a view is displayed or not, if it's not displayed because of a scroll, for me, the test should fail. We want to be happy and friendly with the interactions, but not with the assertions.

By the way, `magicAssert` is a really magic name...

Sloy:
Hi! After 1.5 months of holidays and work stuff, I'm back to this PR!

I'm going to reset myself and just make the simpler version possible of the discussed changes. There's really no need to experiment with _cool_ Kotlin things here. That way nobody can complain that it's too weird of complex. Muahahaha.

About your comment @rocboronat: this doesn't add any more _magic_ to assertions than we already have. We won't try to scroll or anything when asserting. The assertions magic is much simpler than interactions. In this case it's documented in the method JavaDoc:
```
* Attempts to assert using multiple scenarios for the [viewMatcher]:
 * 1. Just one view matches the [viewMatcher].
 * 2. Multiple views match the [viewMatcher]: will pass if at least one of them matches the [condition].
```
Which is basically what we did for https://github.com/SchibstedSpain/Barista/issues/37, https://github.com/SchibstedSpain/Barista/issues/68 or https://github.com/SchibstedSpain/Barista/issues/46.

What this PR does is apply those same conditions to the rest of assertions and keep the code in one place, instead opening a new issue for each assertion that doesn't work as expected and duplicating the code.

The name `magicAssert` is really just a placeholder because I'm not sure how to call it, and wanted to discuss the implementation first. But suggestions are welcomed, because **really** I don't know how to call it. We always talk about "Barista does some magic for you", and we even have a Readme section explaining it. But explaining it in a method name it's not that easy...

Sloy:
I only kept one extension function:
```kotlin
withId(R.id.button).magicAssert(not(isDisplayed()))
```

I think it's the more agreeable one. You guys are ok with that?

Now, another topic: **the naming**. As @rocboronat says, "magicAssert" is a very magic name, which is usually bad. What's the alternative?

ivngrz:
It would be cool to think of ways to use the Kotlin stuff to make the tests more readable, but I agree we should probably discuss that in another PR.

And yep, it would be better to find a name that actually describes what the method is doing 😆  

How about `assertAny`, or `assertAtLeastOnce`? Would that cover it?

We could also go for something more generic, that just says this "not as strict as a default Espresso check", like `softAssert`, or `lenientAssert` or something along those lines.

Sloy:
I like `assertAny`. I think that's a pragmatic description of that the method does.
`withText("button").assertAny(isDisplayed())`.

SmasSive:
Go for `assertAny`!

Sloy:
❗️❗️❗️ Ok, this is the final proposal❗️❗️❗️
 
I updated the code, the title and the description.

Any last complains? Otherwise, let's merge it!

Sloy:
Thanks everyone!

Issue #204, opened 2018-01-20
Better Kotlin support


LPFinch:
Please simplify BaristaRule
Some example
```kotlin
inline fun <reified T : Activity> create(): BaristaRule<T> = BaristaRule.create(T::class.java)
```
After that code looks like 

```kotlin
@Rule
val rule = create<AwesomeActivity>()
```

alorma:
Closed on https://github.com/SchibstedSpain/Barista/pull/205

Thanks!

Sloy:
Thanks @LPFinch for the idea and @alorma for the PR :)

PR #205, opened 2018-01-22
Better kotlin support

alorma:
As requested by @LPFinch, I 've added the support for kotlin on `BaristaRule`

https://github.com/SchibstedSpain/Barista/issues/204

screen shot 2018-01-22 at 11 30 39


Sloy:
Hi @alorma, thanks a lot for the proactivity!

Even though many parts of Barista are written in Kotlin, we want it to be 100% compatible and easy to use from Java. That's why we keep our tests in Java and not Kotlin, to make sure we don't break interoperability with Java.
Would you mind reverting the PermissonGranterTest back to Java?

alorma:
Is the only test that uses BaristaRule, so that's why i decided to change it. We should have a test for this new method

alorma:
Is the only test that uses BaristaRule, so that's why i decided to change it. We should have a test for this new method

alorma:
Rollback on `PermissionGranterTest` @Sloy

SmasSive:
Maybe you should create a new test instead of using an existing one?

README.md
```diff
@@ -294,6 +294,11 @@ All previous rules can be added at the same time by just adding the BaristaRule.
 @Rule
 public BaristaRule<MyActivity> baristaRule = BaristaRule.create(MyActivity.class);
 
+or kotlin way...
```
Sloy: Hmm we don't yet announce or officially support Kotlin APIs. 
Adding this to the readme is a declaration of intent that I'm not sure we want to make... 🤔

library/src/main/java/com/schibsted/spain/barista/rule/BaristaRule.kt
```diff
@@ -0,0 +1,63 @@
...
+class BaristaRule<T : Activity> private constructor(activityClass: Class<T>) : TestRule {
+
+    private val clearPreferencesRule: ClearPreferencesRule = ClearPreferencesRule()
+    private val clearDatabaseRule: ClearDatabaseRule = ClearDatabaseRule()
+    private val clearFilesRule: ClearFilesRule = ClearFilesRule()
+    private val flakyTestRule: FlakyTestRule
+    private val activityTestRule: ActivityTestRule<T>
+
+    init {
+        this.flakyTestRule = FlakyTestRule()
```
Sloy: Why don't just join these initialisations with the declaration? Just like the other ones

library/src/main/java/com/schibsted/spain/barista/rule/BaristaRule.kt
```diff
@@ -0,0 +1,63 @@
...
+    fun launchActivity() {
+        activityTestRule.launchActivity(null)
+    }
+
+    fun launchActivity(startIntent: Intent) {
+        activityTestRule.launchActivity(startIntent)
+    }
+
+    companion object {
+        private val DEFAULT_FLAKY_ATTEMPTS = 10
```
Sloy: The IDE suggested to make these `const`. It's a good suggestion in my opinion.

library/src/main/java/com/schibsted/spain/barista/rule/BaristaRule.kt
```diff
@@ -0,0 +1,63 @@
...
+        private val LAUNCH_ACTIVITY_AUTOMATICALLY = false
+        private val INITIAL_TOUCH_MODE_ENABLED = true
+
+        @JvmStatic
```
Sloy: Since this function is Kotlin-only, I don't think `@JvmStatic` is needed.

alorma:
Done @Sloy

PR #209, opened 2018-02-27
Made public:

gopito:
We have a lot of complex selectors. Lots of "allOf(...)" in our tests on java. So it will be very convenient to have additional public methods with matchers as argument.

Made public:
assertDisplayed(viewMatcher: Matcher<View>)
assertNotDisplayed(viewMatcher: Matcher<View>)
assertNotExist(viewMatcher: Matcher<View>)
clickOn(viewMatcher: Matcher<View>)
scrollTo(matcher: Matcher<View>)
safelyScrollTo(matcher: Matcher<View>)

Added:
assertEffectivelyVisible(viewMatcher: Matcher<View>)

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaVisibilityAssertions.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaVisibilityAssertions.kt
@@ -2,11 +2,11 @@ package com.schibsted.spain.barista.assertion
 
 import android.support.annotation.ColorRes
 import android.support.annotation.IdRes
-import android.support.test.espresso.AmbiguousViewMatcherException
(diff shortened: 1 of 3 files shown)


Sloy:
Hi! Thanks for the contribution @gopito.

In the case of the assert methods, do you think the new method added in https://github.com/SchibstedSpain/Barista/pull/198 will be enough for your use case?

It basically adds a public `assertAnyView(view, condition)` so you can use any matchers you want. Every other assertion method in Barista will be using this method.

Sloy:
I'm glad it helps you.

I just triggered the release of v2.4.0 with the new change. It should be available in a few minutes.

Thanks for reporting you case and contributing :)

Issue #211, opened 2018-03-12
Check if ImageView is displaying a Bitmap


nmwilk:
To test display of images that are not in resources. E.g. when an app downloads content and loads from file.

alorma:
Hi @nmwilk 

You can use `assertHasAnyDrawable(R.id.image_view);` to check if a given ImageView has a drawable

nmwilk:
Hi @alorma. Thanks, but I want to assert that a specific Bitmap is displayed in an ImageView.

Sloy:
Closing it, since we didn't find a solution that works in a deterministic way. 

Manual checks can still be done with https://github.com/SchibstedSpain/Barista/pull/254

PR #212, opened 2018-03-12
Added View assertion for ImageView showing a Bitmap, with a test.

nmwilk:
Implements https://github.com/SchibstedSpain/Barista/issues/211

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaImageViewAssertions.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaImageViewAssertions.kt
@@ -1,14 +1,21 @@
 package com.schibsted.spain.barista.assertion
 
+import android.graphics.Bitmap
 import android.support.annotation.DrawableRes
 import android.support.annotation.IdRes
 import android.support.test.espresso.matcher.ViewMatchers.withId
 import com.schibsted.spain.barista.internal.assertAny
+import com.schibsted.spain.barista.internal.matcher.BitmapDrawableMatcher.Companion.withBitmap
 import com.schibsted.spain.barista.internal.matcher.DrawableMatcher.Companion.withAnyDrawable
 import com.schibsted.spain.barista.internal.matcher.DrawableMatcher.Companion.withDrawable
 import com.schibsted.spain.barista.internal.matcher.DrawableMatcher.Companion.withoutDrawable
 
 object BaristaImageViewAssertions {
+
+    @JvmStatic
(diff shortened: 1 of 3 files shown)


nmwilk:
Ah, I missed that pattern in the tests. Change made as requested. Thanks.

rocboronat:
Wep, just for the information: it should be great to also offer a Bitmap version of `assertHasBackground`. Going to create an issue. Remove it if anyone thinks it's not needed.

Sloy:
Hi! Sorry for being late to the party! We've been away from the project a few months but we're back 💪 

About this feature, I'm worried this might only work for BitmapDrawables and not for images loaded from network or similar. I just tried the matcher by loading the image with Glide and it doesn't work 😞  
`Glide.with(this).load(R.drawable.ic_barista).into(imageView)`

I guess it's because this kind of libraries do transformations on the actual bitmaps, so the expected bitmap will be different from the one loaded into the view.

Not sure how to proceed here 😕

rocboronat:
@Sloy what about explaining it in the error launched by the assertion? You explained so well that all developers should understand it.

rocboronat:
@Sloy the issue is that the methods we're offering right now, `assertHasAnyDrawable(viewId)` and 
`assertHasDrawable(viewId, drawableResId);` are really broad, and in addition, they only work with resources. If you want to test that a given view has a given Bitmap, you can't do it. It's not a big deal for me, because `assertHasAnyDrawable(viewId)` gives a reasonable result, but it's not the ideal one.

In addition, related to our camera intent fake, we could do something to know the result of our mocked camera intent before actually launching it, to write tests to check that the result of a given camera intent actually appears to a given `ImageView`. Right now is a random Bitmap, but it could use a seed or something like that, to enable this kind of tests.

Last words: I think it's not prioritary right now. But let's ask the person who actually invested time on it... @nmwilk is the method `assertHasAnyDrawable()` good enough for you? If not, may you explain why? As we didn't try it in a real environment, we have only the theoric picture of the whole problem... Thanks a lot!

nmwilk:
I don't actually recall the case I was using it for, and I don't have access to that repo anymore. It may have been a test that checks changing an image. E.g. 1/ you choose an image, 2/ the app displays it, 3/ you then choose another image. The test validates step 3 has the new image in.

nmwilk:
So in that case it was working for non-bundled images.

Sloy:
I'm closing this, since we can't be sure it will work on most scenarios. Anyways, this check can also be done with the new `assertAny()` if you need it.

Sloy:
But still, thanks for contributing @nmwilk!

PR #215, opened 2018-04-18
Feature/make public get activity test rule from Barista Rule

denysnovoascmspain:
make public get `activityTestRule ` of the `BaristaRule`, so you can check if the activity is isFinishing

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaAssertions.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaAssertions.kt
@@ -1,12 +1,10 @@
 package com.schibsted.spain.barista.assertion
 
-import android.app.Activity
 import android.support.test.espresso.Espresso.onView
 import android.support.test.espresso.NoActivityResumedException
 import android.support.test.espresso.action.ViewActions
 import android.support.test.espresso.matcher.ViewMatchers.isRoot
 import com.schibsted.spain.barista.internal.failurehandler.RethrowingFailureHandler
-import org.junit.Assert.assertTrue
 import org.junit.Assert.fail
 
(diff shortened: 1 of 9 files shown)


library/src/main/java/com/schibsted/spain/barista/assertion/BaristaAssertions.kt
```diff
@@ -23,4 +25,8 @@ object BaristaAssertions {
         }
     }
 
+    @JvmStatic
+    fun assertActivityIsFinishing(activity: Activity){
```
Sloy: Hmm this feels like a new feature different from what the PRs says.
I'm not sure it adds much value, since it's just a 1-line method for something the activity instance already has. I mean, it's not hiding any complex operations with a better name, it would be OK to just write `assertTrue(activity.isFinishing)` in a real test IMHO. Don't you think?

alorma: Maybe a extension function of barista rule?

`rule.assertActivityIsFinishing()` @Sloy ?

Sloy:
Woah @denysnovoascmspain! Sorry for the delay (a.k.a. I totally forgot).

I made a small change in the test. I'm gonna merge this ASAP

library/src/main/java/com/schibsted/spain/barista/assertion/BaristaAssertions.kt
Sloy: I don't think Barista adds any value  there. We provide shortcuts for things with a complex syntax, like most Espresso methods. But since this doesn't use espresso synchronization or anything, can be asserted with any simple assertion library like junit, assertJ, truth...

PR #222, opened 2018-06-22
Update libraries, build tools, gradle version, ...

alorma:
I've downloaded the project... and it not compiled!!!!!

So here's a fix to that problem, and some update all around


Sloy:
Oh the Travis build fails :(

alorma:
travis is bad, travis hates us

Issue #22, opened 2017-02-23
Scroll to a R.string.text


rocboronat:
BaristaScrollActions.scrollTo(R.string.text) doesn't have any implementation

rocboronat:
Bump! We need this to make this test locale independent:

![image](https://user-images.githubusercontent.com/456499/33031956-0e2bd802-ce20-11e7-9f5c-5d74a6b81ad0.png)

rocboronat:
Barista 2.0 lost this method 😢

Sloy:
😮  But but but but... https://github.com/SchibstedSpain/Barista/pull/169 What happened?

rocboronat:
Don't know :|

rocboronat:
Don't worry, working on it right now

rocboronat:
Done at #176 !

PR #223, opened 2018-06-24
Added the following enhancements:

DeanPike:
1. clickOn function that takes a view id and some text.
2. clearText function that removes the text from a text view.
3. setTimeOnPicker funtion that populates the hour and minute on a time picker.
4. Various methods that will keep checking if a view is displayed until either the view is found or the timeout is reached. Usefull if the view takes a bit of time to render.
5. Updated the assertHint function so that it now handles TextInputLayout and TextInputEditText
6. Added assertError function that will check the error on a TextView or TextInputLayout.
7. Added assertDisplayedAtPosition that will check if a view is displayed at a specified position in a recycler view.


DeanPike:
HI.
Great library.
 I have added some enhancements that I have used in some of my own projects.

README.md
```diff
@@ -52,6 +52,7 @@ Barista already includes `espresso-core` and `espresso-contrib`. If you need [an
 clickOn(R.id.button);
 clickOn(R.string.button_text);
 clickOn("Next");
+clickOn(R.id.button, "Next");
```
Sloy: Why would you need this method when you already have one for ids and another for strings?

README.md
```diff
@@ -126,6 +129,14 @@ assertDisplayed(R.string.hello_world);
 assertDisplayed(R.id.button);
 assertDisplayed(R.id.button, "Hello world")
 
+// Will keep checking until either the view is found or the timeout duration has elaspsed
+assertDisplayedAfterWait(R.id.button);
```
Sloy: I'm not sure about this. It goes against Espresso way of waiting for things, which is by using IdlingResources. You should not manually need to wait for a view or use sleeps.

README.md
rocboronat: It's quite strange to have different Views sharing the same ID but not the same text, at least in our projects 😅
@DeanPike do you have any good example when this `clickOn` is needed or useful? Sometimes, we are tempted to offer this kind of smart and complementary methods, but after writing them, we realize that they are not as useful as we thought and we remove them from the public API.
We try to offer a small API because it's just less code to understand as a user, and less code to maintain as a committer.

rocboronat:
@DeanPike thanks for your contribution! You made a great work, and that work makes this PR very big...
May I request you to split this PR into several ones, so one single PR will represent only one added feature? We totally understand that it's not an easy task, but it will make reviewing this PR a lot easier for us. Right now, if we have something to talk about one of the seven new features, there are six blocked features waiting for the seventh one...

sample/src/main/java/com/schibsted/spain/barista/sample/RecyclerViewActivity.java
```diff
@@ -0,0 +1,69 @@
...
+import android.support.v7.widget.RecyclerView;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.TextView;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class RecyclerViewActivity extends AppCompatActivity {
```
rocboronat: Instead of adding this class, you should use existing `ListsActivity`.

sample/src/main/AndroidManifest.xml
```diff
@@ -1,73 +1,72 @@
...
-    <activity android:name=".KeyboardActivity" />
+        <activity android:name=".RuntimePermissionActivity" />
+        <activity android:name=".SeekBarActivity" />
+        <activity android:name=".SomeSeekBarsWithDifferentProgressesActivity" />
+        <activity
+            android:name=".ColorsActivity"
+            android:exported="true" />
+        <activity android:name=".KeyboardActivity" />
+        <activity android:name=".HintAndErrorActivity" />
+        <activity android:name=".RecyclerViewActivity" />
```
rocboronat: Instead of adding this class, you should use existing `ListsActivity`.

sample/src/main/java/com/schibsted/spain/barista/sample/DatePickerActivity.java
```diff
@@ -2,12 +2,16 @@
...
 import android.support.v4.app.DialogFragment;
 import android.support.v7.app.AppCompatActivity;
 import android.view.View;
 import android.widget.DatePicker;
 import android.widget.TextView;
+import android.widget.TimePicker;
+
 import java.util.Calendar;
 
 public class DatePickerActivity extends AppCompatActivity {
```
rocboronat: Change this name. It shouldn't be `DatePickerActivity` anymore, because now it will also have a `TimePicker`. What about `PickersActivity`? Oh! And change also the layout's name. `activity_pickers`?

sample/src/main/java/com/schibsted/spain/barista/sample/HintAndErrorActivity.java
```diff
@@ -0,0 +1,53 @@
...
+      @Override
+      public void onClick(View view) {
+        showEditTextError();
+      }
+    });
+  }
+
+  private void showLayoutError() {
+    TextInputLayout layout = (TextInputLayout) findViewById(R.id.hintanderror_inputlayout1);
+    layout.setError(getResources().getString(R.string.hintanderror_layout_error));
```
rocboronat: As this code is written inside an `Activity`, you don't need to call `getResources()` to access `getString()`. You may call `getString()` directly.

sample/src/androidTest/java/com/schibsted/spain/barista/sample/RecyclerViewMatcherTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static com.schibsted.spain.barista.assertion.BaristaRecyclerViewAssertions.assertDisplayedAtPosition;
+import static com.schibsted.spain.barista.assertion.BaristaRecyclerViewAssertions.assertRecyclerViewItemCount;
+
+@RunWith(AndroidJUnit4.class)
+public class RecyclerViewMatcherTest {
```
rocboronat: Thanks for those tests!

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaEditTextInteractions.kt
```diff
@@ -14,4 +15,11 @@ object BaristaEditTextInteractions {
         safelyScrollTo(editTextId)
         onView(displayedAnd(withId(editTextId))).perform(replaceText(text))
     }
+
+    @JvmStatic
+    fun clearText(@IdRes viewId: Int) {
```
rocboronat: Great!

library/src/main/java/com/schibsted/spain/barista/assertion/BaristaHintAssertions.kt
```diff
@@ -2,20 +2,49 @@ package com.schibsted.spain.barista.assertion
...
+import org.hamcrest.Matcher
+import org.hamcrest.TypeSafeMatcher
 
 object BaristaHintAssertions {
 
     @JvmStatic
     fun assertHint(@IdRes viewId: Int, @StringRes text: Int) {
-        withId(viewId).assertAny(withHint(text))
+        val string = InstrumentationRegistry.getTargetContext().resources.getString(text)
+        withId(viewId).assertAny(withLayoutHint(string))
```
rocboronat: I propose to call `assertHint(int, String)` instead of doing the actual Espresso call. It will make it easier to spot that it's the same call to Espresso, but this method just easies using a `StringRes` instead of a text.

library/src/main/java/com/schibsted/spain/barista/assertion/BaristaHintAssertions.kt
```diff
@@ -2,20 +2,49 @@ package com.schibsted.spain.barista.assertion
...
     }
 
     @JvmStatic
     fun assertHint(@IdRes viewId: Int, text: String) {
-        withId(viewId).assertAny(withHint(text))
+        withId(viewId).assertAny(withLayoutHint(text))
     }
 
+    private fun withLayoutHint(expectedHint: String): Matcher<View> {
```
rocboronat: I think we may think in a better name for this method... but I can't propose any 😅

library/src/main/java/com/schibsted/spain/barista/assertion/BaristaErrorAssertions.kt
```diff
@@ -0,0 +1,47 @@
...
+import android.support.test.InstrumentationRegistry
+import android.support.test.espresso.matcher.ViewMatchers
+import android.view.View
+import android.widget.TextView
+import com.schibsted.spain.barista.internal.assertAny
+import org.hamcrest.Description
+import org.hamcrest.Matcher
+import org.hamcrest.TypeSafeMatcher
+
+object BaristaErrorAssertions {
```
rocboronat: I have the same comments here than in the similar class `BaristaHintAssertions`

DeanPike:
Hi Rafa.

I agree with the your comment on the "clickOn(R.id.button, "Next");"
method.
Just thought it might be useful. I am happy to remove it.

Regarding the "assertDisplayedAfterWait" we have found that some of our
jenkins builds fail randomly and it seemed to be that the test was running
before the view rendering had completed,.
Putting a delay into that assert seemed to sort out the issue.

I will have another look at the Espresso IdlingResources and see If I can
remove this.

DeanPike:
Hi

I just thought it might be a nice to have but am happy to remove it.

DeanPike:
Not a problem.
The change started like that but got away from me in the end.

I will create pull requests for the individual changes.

Sloy:
About the clickOn function, I've been talking internally with people from different teams, and we agree that there are many edge cases that are not covered by the library. Barista usually provides two versions of its methods: with id or with text. Adding new combinations for each special case would flood the library with too many methods.

Having said that, I'm thinking we could just provide a third version of the methods that receives a `Matcher<View>`, to allow people to use custom matching rules for their own cases. I created an issue to discuss it. Feel free to add your opinion there :) https://github.com/SchibstedSpain/Barista/issues/22

PR #226, opened 2018-06-28
Added preliminar dokka config

Sloy:
This allows generating java or kotlin documentation with the `dokka` gradle task.
I added config to exclude our internal packages, so the documentation only reflects the public API.

It can be viewed by running the task `./gradle dokka`. The files are generated under `library/build/javadoc`. 

Now the question is: Is it useful? Do we want it?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,6 +8,7 @@ buildscript {
         classpath 'com.android.tools.build:gradle:3.1.3'
         classpath 'com.novoda:bintray-release:0.8.1'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+        classpath "org.jetbrains.dokka:dokka-android-gradle-plugin:0.9.17"
     }
 }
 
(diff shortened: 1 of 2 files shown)


rocboronat:
Hi! I like this PR because it's an easy way to get introduced to `dokka`. It was totally unknown for me, thanks for the introduction!

By the way, I think that right now, the API is easy enough. I mean, what will be the Javadoc for the `clickOn(id)` method? I'm afraid of having lots of code explaining things that should not be explained, like the old Javadoc autogenerators did some years ago: `clickOn(id)` _Clicks on the View with the passed ID_.

I don't have a strong opinion about if we should apply this PR or not. Probably we should ask to Barista's newcomers: Do you think that Javadoc will make using Barista an easier task?

rocboronat:
By the way... right now, the library has Javadocs that explains things to other Barista developers, not the actual Barista users. We should write a great documentation if we apply this PR :·)

`Scrolls in Espresso are not as great as we could expect. For that reason, we will try to scroll several times instead of just one.` Uh.

SmasSive:
Nope, go ahead!

PR #227, opened 2018-06-29
Additional enhancements

DeanPike:
Hi.

I have separated the changes into individual commits and made the suggested chages.


rocboronat:
Ohhh! The idea was to separate it in different PR's 😅For example: one PR for the `clearText()` method, another one with the tests you added about the lists... Doing this is great, because if something is easy to merge (like the two examples I have given), we will merge it in no time 👍
I'm sorry for requesting more and more work...

DeanPike:
Not a problem. I can split them up.

PR #234, opened 2018-07-01
Added clearText function that removes the text from a text view.

DeanPike:


sample/src/androidTest/java/com/schibsted/spain/barista/sample/EditTextTest.java
```diff
@@ -39,12 +41,11 @@ public void checkWriteOnEditText_whenParentIsNotAScrollView() {
   }
 
   @Test
-  public void assertHintById() {
```
Sloy: Watch out! You just removed a couple of existing tests 😬

DeanPike:
Oops.
My bad.
Will add them back and re-submit.

Thanks

PR #235, opened 2018-07-02
Recycler view enhancements

DeanPike:


Sloy:
The PR accidentally included commits from other PRs. I removed those commits and only keep the changes related to the RecyclerView.

README.md
```diff
@@ -207,6 +207,8 @@ assertTextColorIsNot(R.id.someColorListText, R.color.another_state_list);
 
 // Check recyclerView item count against expected item count
 assertRecyclerViewItemCount(R.id.recycler, 10);
+assertDisplayedAtPosition(R.id.recycler, 0, "text");
+assertDisplayedAtPosition(R.id.recycler, 0, R.id.text_field, "text");
```
Sloy: I have doubts about the name. This method only works for recycler views. But that's not very clear from the name.

I wonder if we should me more explicit, like `assertRecyclerItemDisplayedAtPosition()`. Or `assertDisplayedAtListPosition()`.
Or maybe I'm thinking too much and it's not really a problem. In the future we might improve it to make it compatible with ListViews too, like we did for `BaristaListInteractions`.

Opinions @SchibstedSpain/barista-mantainers?

alorma: Agree with @Sloy 

Even more, adding a list id, position, view id and text is.... too much, is not easily to remember that method and the params

Maybe we can think on something more simplier, or try to reduce the number o params with helper class? not sure..

library/src/main/java/com/schibsted/spain/barista/assertion/BaristaRecyclerViewAssertions.kt
```diff
@@ -1,15 +1,92 @@
...
+        .perform(scrollToPosition<RecyclerView.ViewHolder>(position))
+
+    onView(atPositionOnView(recyclerViewId, position, viewId))
+        .check(matches(withText(text)))
+  }
+
+  /**
+   * Fork of the RecyclerViewMatcher from https://gist.github.com/baconpat/8405a88d04bd1942eb5e430d33e4faa2
+   */
+  private fun atPosition(@IdRes recyclerViewId: Int, position: Int): Matcher<View> = atPositionOnView(recyclerViewId, position, -1)
```
SmasSive: I'd suggest to use default params here so you'll have only one method `atPosition` with the default value of the `viewId = -1`. In addition I'd extract that `-1` to a constant named for example `NO_VIEW_ID`.
If you do that, you'll be able to do the same for `assertDisplayedAtPosition` and you'll simplify the code.

README.md
SmasSive: `assertRecyclerItemDisplayedAtPosition` sounds good to me 👍

alorma: `assertTextDisplayedAtPosition` sounds good? ???

rocboronat: My proposal is
`assertListPositionHasText(recyclerId, position, String)`
`assertListPositionHasText(recyclerId, position, String, textViewId)`
`assertListPositionHasText(recyclerId, position, textRes)`
`assertListPositionHasText(recyclerId, position, textRes, textViewId)`

By the way, @DeanPike, please swear that you really need those methods 😅

sample/src/androidTest/java/com/schibsted/spain/barista/sample/RecyclerViewMatcherTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static com.schibsted.spain.barista.assertion.BaristaRecyclerViewAssertions.assertDisplayedAtPosition;
+import static com.schibsted.spain.barista.assertion.BaristaRecyclerViewAssertions.assertRecyclerViewItemCount;
+
+@RunWith(AndroidJUnit4.class)
+public class RecyclerViewMatcherTest {
```
rocboronat: We should add some tests in this class to assert that the assertions fail if the condition isn't met.

README.md
DeanPike: Hi Roc.

I do need this functionality as I`m using it in a project.

DeanPike:
Hi.

I have moved the recycler view asserts into the BaristaListInteractions class.
The functions now work with recycler and list views.

Issue #239, opened 2018-07-02
Better handling of failures


Sloy:
Until now we didn't take good care of how Barista fails. We have some tests to check that it fails, but not how it fails.

After looking around in the code I found out some problems with failures: Sometimes we notify Espresso's failure handler, some times we don't (#220), sometimes we throw a BaristaException with a nice message, sometimes we just throw the default Espresso AssertionError, sometimes the error messages don't give any hint about the real issue.

It's a mess.

I'm going to try to fix that incrementally. I'm opening this issue to serve as a common reference for the PRs I will open.

radixdiaboli:
The thing is, I'm not willing to put much work into this (including filing issues) without first establishing what your failure policy is to begin with.  And this issue seems like the place to discuss that, no?

radixdiaboli:
"By the way, I don't agree you. If PermissionGranter doesn't work and people still has green tests, it means that their tests don't need any permission, right?"  

The example I gave actually shows that it doesn't mean that.  The code fails to press the "Allow" button to grant a permission, and yet the test still passes.  That's a _huge_ issue.

Additionally, the user might also have chosen  the option to not be asked for permission anymore, and so the dialog isn't showing.

Even if you already had the permission and that's why the dialog isn't showing, it would still be bad for this to pass.  If I were to write a test that said "Click(r.id.button)" and there was no button, that should throw a fail. If I were to click a notification and it couldn't click it, that should fail. Ditto for every other action you might want to emulate. If you're thinking of permissions as just a system level thing for your tests to deal with when they happen to come up, you're doing it wrong.

That's because every test is an explicit *expectation* of what is going to happen in the app.  If I'm expecting to need a permission and it doesn't come up, then my expectation was wrong the test should fail to let me know.  What I should *not* do is write flaky test code to gloss it over, which is what this function is.

Even if we disagree about whether these test should fail, the way it is currently written does not give me any option.  If it did throw an exception, the calling function could always wrap it with a try/catch block to ignore it.  But the reverse isn't true: I can't do anything to make this function fail to pass.  I guess you can't unflake a flaky function? :-)

So, using this as a case study in fail policy, I think the policy should be to be consistent with Espresso: if some action does not success, throw a failure, let the user decide what to do.  The alternative is inevitably false positives, and gives the user less options with how to write their tests.

rocboronat:
Hi again!

It seems that you're forgetting that Barista runs on test environments. The user never taps the "don't ask me again" button, for example.

[...]

PR #240, opened 2018-07-02
Fix ListInteractions failure handling

Sloy:
The error handling of clickListItem and clickListItemChild were not great. This is part of https://github.com/SchibstedSpain/Barista/issues/239

Basically:
1. We were not notifying Espresso's FailureHandler when the methods fail
2. The exception messages don't provide good info about what went wrong.

I improved the tests with AssertJ's `catchThrowable()`, and a new `SpyFailureHandlerRule` similar to `FailureHandlerValidatorRule` but with less unpredictable magic.

This way we make sure that 
1. The Espresso's FailureHandler is invoked only once or never, depending on the test case.
2. The Exception message gives the information we expect, and we don't accidentally break it in the future.

PS: I added Java 8 support to the sample project, so we can use lambdas in the tests for:
```java
Throwable thrown = catchThrowable(() -> clickListItem(999));
assertThat(thrown).isInstanceOf(BaristaException.class)
    .hasMessageContaining("Could not perform action")
```
Thanks @jaumejane90 and @javiyt!

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/interaction/BaristaListInteractions.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/interaction/BaristaListInteractions.kt
@@ -17,11 +17,14 @@ import android.support.test.espresso.matcher.ViewMatchers.isDisplayed
 import android.support.test.espresso.matcher.ViewMatchers.withId
 import android.support.v7.widget.RecyclerView
 import android.support.v7.widget.RecyclerView.ViewHolder
+import android.view.View
 import android.widget.AbsListView
 import com.schibsted.spain.barista.internal.failurehandler.SpyFailureHandler
+import com.schibsted.spain.barista.internal.failurehandler.description
 import com.schibsted.spain.barista.internal.failurehandler.withFailureHandler
 import com.schibsted.spain.barista.internal.viewaction.ClickChildAction.clickChildWithId
 import com.schibsted.spain.barista.internal.viewaction.PerformClickAction.clickUsingPerformClick
+import org.hamcrest.Matcher
 import org.hamcrest.Matchers.allOf
 import org.hamcrest.Matchers.anyOf
 import org.hamcrest.Matchers.anything
@@ -58,11 +61,19 @@ object BaristaListInteractions {
     private fun performMagicAction(@IdRes id: Int?, position: Int, recyclerAction: ViewAction, listViewAction: ViewAction) {
         verifyOneSingleMatch(id)
 
+        val recyclerMatcher = findRecyclerMatcher(id)
+        val listViewMatcher = findListViewMatcher(id)
         val spyFailureHandler = SpyFailureHandler()
         try {
-            performOnRecycler(id, recyclerAction, spyFailureHandler)
+            performOnRecycler(recyclerMatcher, recyclerAction, spyFailureHandler)
         } catch (noRecyclerMatching: NoMatchingViewException) {
-            performOnListView(id, position, listViewAction, spyFailureHandler)
+          try {
+            performOnListView(listViewMatcher, position, listViewAction, spyFailureHandler)
+          } catch (listViewError: Throwable) {
+            spyFailureHandler.resendLastError("Could not perform action (${listViewAction.description}) on ListView ${listViewMatcher.description()}")
+          }
+        } catch (recyclerError: Throwable) {
+            spyFailureHandler.resendLastError("Could not perform action (${recyclerAction.description}) on RecyclerView ${recyclerMatcher.description()}")
         }
     }
 
@@ -84,16 +95,16 @@ object BaristaListInteractions {
         }
     }
 
-    private fun performOnRecycler(@IdRes id: Int?, recyclerAction: ViewAction, failureHandler: FailureHandler) {
-        onView(findRecyclerMatcher(id))
+    private fun performOnRecycler(matcher: Matcher<View>, recyclerAction: ViewAction, failureHandler: FailureHandler) {
+        onView(matcher)
                 .withFailureHandler(failureHandler)
                 .perform(recyclerAction)
     }
 
-    private fun performOnListView(@IdRes id: Int?, position: Int, listViewAction: ViewAction, failureHandler: FailureHandler) {
+    private fun performOnListView(matcher: Matcher<View>, position: Int, listViewAction: ViewAction, failureHandler: FailureHandler) {
         withFailureHandler(failureHandler) {
             onData(anything())
-                    .inAdapterView(findListViewMatcher(id))
+                    .inAdapterView(matcher)
                     .atPosition(position)
                     .perform(listViewAction)
         }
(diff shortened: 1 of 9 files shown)


sample/src/androidTest/java/com/schibsted/spain/barista/sample/ListsChildClickTest.java
```diff
@@ -23,94 +26,109 @@
...
-  public void fails_whenListViewChildNotExist() throws Exception {
+  @Test
+  public void fails_whenRecyclerChildNotExist() {
     openActivity(ListsActivity.buildIntent()
-        .withComplexLists(R.id.listview)
+        .withRecyclers(R.id.recycler)
     );
 
-    clickListItemChild(20, R.id.unknown);
+    Throwable thrown = catchThrowable(() -> clickListItemChild(20, R.id.unknown));
```
SmasSive: Cool!

sample/src/androidTest/java/com/schibsted/spain/barista/sample/ListsChildClickTest.java
```diff
@@ -23,94 +26,109 @@
...
 
-    assertResult("");
+    spyFailureHandlerRule.assertEspressoFailures(1);
+    assertThat(thrown).isInstanceOf(BaristaException.class)
+        .hasMessageContaining("Could not perform action (actionOnItemAtPosition performing ViewAction: Click on a child view ")
```
SmasSive: Doesn't it is a little too flaky?

sample/src/androidTest/java/com/schibsted/spain/barista/sample/util/SpyFailureHandlerRule.kt
```diff
@@ -0,0 +1,47 @@
...
+    return object : Statement() {
+      override fun evaluate() {
+        Espresso.setFailureHandler(failureHandler)
+        try {
+          base.evaluate()
+        } catch (t: Throwable) {
+          failed = true
+          throw t
+        } finally {
+          if (!failed && !asserted) {
```
SmasSive: Meeeh...

SmasSive: I **don't** like this code at all... I know I should make a constructive criticism and give you an alternative but I don't have it so... Just I wanted to let you know that I'm not happy with it 🤷‍♂️

sample/src/androidTest/java/com/schibsted/spain/barista/sample/ListsChildClickTest.java
Sloy: Yep. Or even if we update Espresso and it changes something in the message format. 
But I think the benefit of having the full message here is greater than the cost. Plus it serves as documentation of the messages from Barista. Previously we had no idea of what it would show, and some of them were actually wrong.

Also, if we see they fail too often we can always change them for more generic checks.

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaListInteractions.kt
```diff
@@ -58,11 +61,19 @@ object BaristaListInteractions {
...
+        val listViewMatcher = findListViewMatcher(id)
         val spyFailureHandler = SpyFailureHandler()
         try {
-            performOnRecycler(id, recyclerAction, spyFailureHandler)
+            performOnRecycler(recyclerMatcher, recyclerAction, spyFailureHandler)
         } catch (noRecyclerMatching: NoMatchingViewException) {
-            performOnListView(id, position, listViewAction, spyFailureHandler)
+          try {
+            performOnListView(listViewMatcher, position, listViewAction, spyFailureHandler)
+          } catch (listViewError: Throwable) {
```
danieldisu: You are not catching the NoMatchingViewException for the listView ""try"" is this intended?

sample/src/androidTest/java/com/schibsted/spain/barista/sample/ListsChildClickTest.java
SmasSive: Ok you got me! 🏆

sample/src/androidTest/java/com/schibsted/spain/barista/sample/util/SpyFailureHandlerRule.kt
Sloy: Then shut up!

Just kidding, if your _spider sense_ is telling you this is bad, please think about what you don't like and how to improve it.
Is it the behavior of the code that you don't like? Or how the implementation looks?
I'm absolutely open to suggestion!

SmasSive: It's the implementation, IMO using flags to control the flow of the code is not a good practice because it's too weak and could end up in an unexpected behaviour.
On the other hand, this is not a key code neither complex and I'm sure it won't fail... In addition it's a test code so, what we should worry about it?
Go ahead and forget my mimimi's!

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaListInteractions.kt
Sloy: Yes it is. The idea behind this magic is: We don't know if you'r using a Recycler or a ListView, so we'll first try to do it on a recycler, and if we don't find it we'll try on a ListView.

And what if it doesn't find a ListView either? That won't happen, because in that case the method will fail before this code, in the `verifyOneSingleMatch()` invocation. 

The implementation is a bit confusing, I know. But at least it's covered by tests!
```java 
  @Test
  public void fail_whenNoViewFound() {
    launchTestActivity(ListsActivity.buildIntent());

    Throwable thrown = catchThrowable(() -> clickListItem(20));

    spyFailureHandlerRule.assertEspressoFailures(1);
    assertThat(thrown).isInstanceOf(BaristaException.class)
        .hasMessage("No ListView or RecyclerView found in the hierarchy");
  }
```

_See @SmasSive? Thanks to the verbose message in the test we know that we're doing it right_ :p

danieldisu: Oh I see, I'm not familiar with the architecture and I thought that something like this would happen!

sample/src/androidTest/java/com/schibsted/spain/barista/sample/util/SpyFailureHandlerRule.kt
Sloy: Done!

Sloy:
With the awesome help of @SmasSive, we've covered the ugly logic with unit tests, and then refactored it to a simpler version. Turns out that @SmasSive's spider sense was right! The code could be much much more simpler and cleaner. But without the tests neither of us thought of that solution.
❤️

Sloy:
No more comments, only emojis. I'll assume y'all are OK with the changes now :D

PR #243, opened 2018-07-07
Safer view pager swipe actions

### LINKED ISSUE #210: Add more options to view pager actions
Hey there,

From Karumi, we've been playing with your library for a while and, first of all, let me say that it's awesome, thank you all for saving us some precious time!

We found some very specific scenarios where the view pager actions didn't work out for us, though.
1. We've been working with view pagers that are not completely visible on the screen. For that case, we had to create our own `GeneralSwipeAction` copy to be able to ignore the visibility requirement (see [the original implementation](https://android.googlesource.com/platform/frameworks/testing/+/android-support-test/espresso/core/src/main/java/android/support/test/espresso/action/GeneralSwipeAction.java#40) where it forces the swipe action to apply iff the 90% of the view is on the screen).
2. The other workaround we came with is to make it possible to change the initial/end position of the gesture itself. The problem being that we had a view pager AND a drawer in the same screen. Starting the movement from the left side of the screen would open the drawer and doing it from the center of the view wouldn't go to the next page when running on slow environments (I'm looking at you, CI  👀).

Would you be interested on adding those features to Barista?

Serchinastico:
### :pushpin: References
* **Issue:** Fixes #210 

###  💻 How is it being implemented?

* Create a copy of `GeneralSwipeAction` called `GeneralSwipeWithPartiallyVisibleViewAction` that does not take care of the target view visibility. The original implementation has a hardcoded value preventing the swipeable view to have at least the 90% of it displayed on the screen. This one just doesn't.
* Create new `swipeRight` and `swipeLeft` actions that start from one of the edges and goes straight to the opposite edge (instead of stopping at the center of the view, like the original implementation does). We've experienced in some (really slow) CI environments that going to the center of the view is not enough and most of the times it just lets the view pager in the same page.

### 🤖 How can it be tested?

* I included automatic tests to see if a view pager with a 3rd of it out of the screen works with the new implementation and added a drawer to make sure we are not opening it by unintentionally.

### 🗒 Notes

* I ended up not including additional options and just replacing the current implementation. In the end, this is everything we needed to address our main problems with view pager interactions.


library/src/main/java/com/schibsted/spain/barista/internal/viewaction/GeneralSwipeWithPartiallyVisibleViewAction.kt
```diff
@@ -0,0 +1,134 @@
...
+import android.support.test.espresso.action.Swipe
+import android.support.test.espresso.action.Swiper
+import android.support.test.espresso.action.ViewActions
+import android.support.test.espresso.matcher.ViewMatchers
+import android.support.test.espresso.util.HumanReadables
+import android.view.View
+import android.view.ViewConfiguration
+import org.hamcrest.Matcher
+
+object SwipeActions {
```
Sloy: Since this file contains two classes, one public and one private; how about naming the file after the public class instead of the private one? 

I.e. naming the file `SwipeActions`.

sample/src/androidTest/java/com/schibsted/spain/barista/sample/PartiallyVisibleViewPagerTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static com.schibsted.spain.barista.assertion.BaristaVisibilityAssertions.assertDisplayed;
+import static com.schibsted.spain.barista.assertion.BaristaVisibilityAssertions.assertNotDisplayed;
+import static com.schibsted.spain.barista.interaction.BaristaViewPagerInteractions.swipeViewPagerBack;
+import static com.schibsted.spain.barista.interaction.BaristaViewPagerInteractions.swipeViewPagerForward;
+
+@RunWith(AndroidJUnit4.class)
+public class PartiallyVisibleViewPagerTest {
```
Sloy: I tried running the tests without the workaround (reverting the swipeLeft and swipeRight imports) to see if I can reproduce the error, and they pass too. 
I think I understand the issue and I agree it should fail in the given scenario. Maybe I'm missing something?

Serchinastico: That's... weird, I did the same verification as you did and I'd swear they failed but let me give it another try

Serchinastico: You were right, I don't know if I changed something in the layout in the last minute but it was working with both implementations. I just put the view pager inside a frame layout and looks like I now can reproduce the behavior that led me to this PR :+1:

Serchinastico:
Updated with your suggestions

PR #245, opened 2018-07-11
#244 Update write methods to work with wrapped EditTexts

### LINKED ISSUE #244: writeToAutoComplete with SearchView (compat) not works
Hi @SchibstedSpain/barista-maintainers !

Recently I found the case where we have the following widget on XML:

```
            <android.support.v7.widget.SearchView
                android:id="@+id/searchview_search"
                android:layout_width="match_parent"
                android:layout_height="wrap_content"
                android:gravity="left"
                android:queryHint="@string/search"/>
```

Using `SearchView` is different that normal `EditText` as it extends from `LinearLayoutCompat`

The only way that i found to use this view was the following (using internal ID):

```
writeToAutoComplete(android.support.v7.appcompat.R.id.search_src_text, "Pro")
```

Would be great to have something like:

```
writeToSearchView(R.id.searchview_search, "Pro")
```

What do you think??

alorma:
> After some discussions with @Sloy the feature had changed a bit...

This feature adds support on `writeTo` and `writeToAutoComplete` for views that has an `EditText` inside, like `SearchView` or `TextInputLayout`

Fix #244


Sloy:
The PR title and description are outdated, right?

alorma:
Updated!

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaAutoCompleteTextViewInteractions.kt
```diff
@@ -1,14 +1,22 @@
...
+import org.hamcrest.Matchers.allOf
+import org.hamcrest.Matchers.anyOf
 
 object BaristaAutoCompleteTextViewInteractions {
 
     @JvmStatic
     fun writeToAutoComplete(@IdRes autoCompleteId: Int, text: String) {
-        withId(autoCompleteId).performAction(replaceAutoComplete(text))
+        val withId = withId(autoCompleteId)
+        val assignableFrom = isAssignableFrom(EditText::class.java)
```
Sloy: Personal opinion: I think that extracting these two matchers to a variable with the same name of the matcher itself only adds noise.
Just writing `val simpleMatcher = allOf(withId(autoCompleteId), isAssignableFrom(EditText::class.java))` seems easier to understand.

Sloy:
Tip: If you write "Fixes #244" in the PR description, GitHub should automatically close the issue when the PR is merged :D

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaAutoCompleteTextViewInteractions.kt
alorma: i will agree if those are used just on `simpleMatcher`, but they are used also on `wrappedMatcher`

PR #250, opened 2018-08-27
Clear files with optional regex filter

Sloy:
Simply deleting all files might break things for some apps.
For example, a library might depend on some files that get stored when initialized only.
Or you might have a set of files that don't want deleted between test.

For that reason I added a regex filter to delete only a set of files.
I tried to mimic the api and implementation of ClearDatabaseRule, and added unit tests.


library/src/main/java/com/schibsted/spain/barista/rule/cleardata/internal/FileOperations.kt
```diff
@@ -0,0 +1,33 @@
...
+  fun getAllFilesRecursively(): Collection<File> {
+    return androidDirectories.flatMap { listFileTree(it) }
+  }
+
+  private fun listFileTree(dir: File?): Collection<File> {
+    if (dir?.listFiles() == null) {
+      return emptySet()
+    }
+    val fileTree = HashSet<File>()
+    dir.listFiles().forEach { entry ->
```
jffiorillo: This is fine but you can also do `return dir.listFiles().map{}`... and not declaring `fileTree` variable.

Sloy: Hmm but it's doing different operations depending if it's a file or a directory. `add` vs `addAll`.
How would you do it with a map?

ferransimon: Maybe something like this: (quite ugly, but functional)

dir.listFiles().stream().map(entry -> {
   if (entry.isFile) {
        return Arrays.asList(entry)
      } else {
        return listFileTree(entry)
      }
}).flatMap(entry -> entry.stream())
.collect(Collectors.toList());

Sloy: I just found out an even nicer way...

```kotlin
return dir.walkTopDown()
        .filter { it.isFile }
        .toList()
```

Kotlin will never stop surprising me 😂

PR #253, opened 2018-09-22
Created class BaristaListAssertions 

DeanPike:
Created class BaristaListAssertions to hold list and recyclerview assertions.
These assertions allow the finding of an item in a recycler or list view


Sloy:
Thanks @DeanPike

PR #254, opened 2018-10-01
Add assertAny {} with sugar for kotlin

alorma:
```kotlin
assertAny<EditText>(R.id.edittext) {
        it.text.toString() == "Hello!"
}
```

Now, we can use this new `assertAny` method in kotlin, and inside the lambda, return a `boolean`:D

This method(s) will help to create custom assert for any view without know how to build *matchers*

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,39 @@ assertProgressIsMax(R.id.seek_bar)
 
 ```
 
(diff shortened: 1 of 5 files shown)


alorma:
Thanks @Sloy updated, does it matches what you expect now?

PR #255, opened 2018-10-04
Improve the explanation about custom assertions

rocboronat:
Hello from South Korea! 🌏

#254 is a great PR, but I needed to read its code twice to understand what it was offering. The readme didn't explain the whole power of the new feature.

This PR will try to better explain it. Feel free to improve it as much as possible. I tried to focus on newbies instead of Espresso experts because that's what I am 🤗

By the way, if I misunderstood the API that was offered, feel free to shame on me 💩and commit on this PR without previous notice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,7 +257,7 @@ assertAny<TextView>(withId(R.id.textView)) {
     it.text.toString() == "Hello!"
 }
 
-// You can also define the assertion error message if the assertion fails
+// You can also define the assertion error message that will be shown if the assertion fails
 assertAny<TextView>(R.id.textView, "text wasn't Hello!") {
     it.text.toString() == "Hello!"
 }
--- a/README.md
+++ b/README.md
@@ -238,33 +238,27 @@ assertProgressIsMax(R.id.seek_bar)
 
 ### Custom assertions
 
-If you have a special case not covered by the given assertions API, we encourage you to assert these special cases with our custom assertions API. It's a convenient way to replace plain `Matcher`s with complex assertions. With Barista, you can match any kind of view by knowing its type and passing its `viewId`, `text`, or a common `Matcher<View>`. Once you matched it, you will be able to assert all its properties without adding any complex `Matcher` to your project.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -236,35 +236,35 @@ assertProgressIsMax(R.id.seek_bar)
 
 ```
 
-### Barista custom assert
+### Custom assertions
 
-Do your own asserts on given view, no more custom *Matcher*
+If you have a special case not covered by the given assertions API, we encourage you to assert these special cases with our custom assertions API. It's a convenient way to replace plain `Matcher`s with complex assertions. With Barista, you can match any kind of view by knowing its type and passing its `viewId`, `text`, or a common `Matcher<View>`. Once you matched it, you will be able to assert all its properties without adding any complex `Matcher` to your project.
```
alorma: `common` or `custom` Matcher?

rocboronat:
@alorma totally! Tomorrow I'll update the PR

README.md
rocboronat: Mmmmm you can also use a common `withId` or those given by Espresso... but you are right. I think I could just avoid the word and it will work better 💃

rocboronat:
@alorma done! I copied them and I think I improved them a bit. Now the examples show all the power of the feature. Let's check again! There's always space for the improvement 🚀🍌

Sloy:
Hi! I'm off this week, but the error message is wrong :)
It's not an error message, it's the "expected" message. Because when the matcher fails Espresso will print "condition didn't match <message>". 
It's not very intuitive, but it's how matchers work. They're called descriptions, not error messages.

rocboronat:
@Sloy totally agree, but we are hiding the complexity and counterintuitive things of the Matchers with our approach, aren't we?
If we aren't, what do you think about changing `assertion error message` by `assertion description message` in the README?

rocboronat:
@Sloy thanks for the clarification! The thing is that `description` is broad enough that I didn't understand its meaning. I thought you were talking about the description of the View, not the assertion.

By the way, going to fix the comment inlined in the code.

Issue #262, opened 2018-10-30
Upgrade SupportLib dependency to AndroidX


sridhar-vadlamani-ck:
AndroidX is the latest and greatest by Google. As 28 is the last support lib version, upgrading to androidx will futureproof this.

Sloy:
The problem with using AndroidX is that we will force library users to update to AndroidX too. Is that right? That's not something we want to do at the moment, not until AndroidX is more widespread.

We haven't migrated to AndroidX internally yet, but if I'm not mistaken the Jetifier tool allows apps to use libraries consuming the support library. So this should not be a blocking issue.

Let me know if I got something wrong.

Sloy:
But let's keep the issue open, because we'll want to do it someday 😄

lachudra87:
Problem with Jetifier and AndroidX in case of barista lib is that `uiautomator-v18` is only in alpha version for AndroidX and it looks like only non v18 version of `uiautomator` is supported. So Jetifier won't override `uiautomator-v18` and there will be duplicated class definition in dex. Solution is to manually exclude `ui-automator-v18` from barista import with sth like below:
```
    androidTestImplementation ('com.schibsted.spain:barista:2.7.0') {
        exclude group: 'com.android.support.test.uiautomator', module: 'uiautomator-v18'
    }
```
But maybe it is possible to change of `uiautomator` implementation in barista itself to make life of developers a little bit easier ;)

DeanPike:
Would it be possible to create a branch of the library that uses androidx so that we can use the latest version of espresso as it looks like the pre androidx versions of espresso are not going to be supported any more.

As of version 3.1.0 Espresso seems to have migrated to androidx libraries.
com.android.support.test.espresso.espresso-core version 3.0.2 was last updated in Apr 2018
androidx.test.espresso.espresso-core version 3.1.1 was last updated in Dec 2018

Sloy:
@DeanPike maybe we should just release a 3.0.0 version of Barista including AndroidX. Everyone using the support lib can still use Barista 2.x until they migrate. They would not get new features though.

What do you think @alorma @rocboronat @sergiocastilloscmspain ?

Sloy:
@lachudra87 hmm let me check that, because I recently migrated a project to AndroidX using Barista and didn't have any problems with uiautomator. 
In any case, we'll try to do as much as possible to make Barista compatible with Jetifier.

mrk-han:
[...]

PR #264, opened 2018-11-08
I have changed the codebase to use androidx.

DeanPike:
In order to do this I had to update the following dependencies.

Changed kotlin to 1.3.0
Changed com.android.tools.build:gradle to 3.2.1
Changed gradle wrapper to user version 4.10.2
Changed support library version to 1.0.0
Changed espresso version to 3.1.0
Changed uiAutomatorVersion version to 2.2.0
Changed minSdkVersionDeclared to 18
Changed the classes path in the android-quality.gradle file


SmasSive:
@DeanPike thanks for a great contribution!

In my opinion this should not happen right now because:
- Migrating to androidx will make Barista incompatible with other versions
- We could do a new module and publish two separate artifacts but it will lead us to duplicate all the code and this will not happen for the sake of our mental health
- Another option could be to deprecate the current version and go for androidx but I think it's too early to do it, I don't know the numbers of adoption of androidx but I bet it's not high enough to give it all

So, it's a good contribution and I think it will happen sooner than later but not now.

alorma:
I'm not sure about this change also, even if it's a great addition and I'm sure it took a some time to do it, we appreciate the effort!

Let's move the discussion to: https://github.com/SchibstedSpain/Barista/issues/262, so we don't mess a PR with discussion (where issues belongs)

Issue #260, opened 2018-10-27
Use top level functions


PaulWoitaschek:
Creating kotlin objects just for the sake of holding functions is not idiomatic.
Instead it's recommended to use top level functions. For java interop, there is `@file:JvmName`.

For example `BaristaImageViewAssertions` would change from:

```kotlin
package com.schibsted.spain.barista.assertion

import android.support.annotation.DrawableRes
import android.support.annotation.IdRes
import android.support.test.espresso.matcher.ViewMatchers.withId
import com.schibsted.spain.barista.internal.assertAny
import com.schibsted.spain.barista.internal.matcher.DrawableMatcher.Companion.withAnyDrawable
import com.schibsted.spain.barista.internal.matcher.DrawableMatcher.Companion.withDrawable
import com.schibsted.spain.barista.internal.matcher.DrawableMatcher.Companion.withoutDrawable

object BaristaImageViewAssertions {

    @JvmStatic
    fun assertHasDrawable(@IdRes imageViewId: Int, @DrawableRes drawable: Int) {
        withId(imageViewId).assertAny(withDrawable(drawable))
    }

    @JvmStatic
    fun assertHasAnyDrawable(@IdRes imageViewId: Int) {
        withId(imageViewId).assertAny(withAnyDrawable())
    }

    @JvmStatic
    fun assertHasNoDrawable(@IdRes imageViewId: Int) {
        withId(imageViewId).assertAny(withoutDrawable())
    }
}
```

To:
```kotlin
@file:JvmName("BaristaImageViewAssertions")

package com.schibsted.spain.barista.assertion

import android.support.annotation.DrawableRes
import android.support.annotation.IdRes
import android.support.test.espresso.matcher.ViewMatchers.withId
import com.schibsted.spain.barista.internal.assertAny
import com.schibsted.spain.barista.internal.matcher.DrawableMatcher.Companion.withAnyDrawable
import com.schibsted.spain.barista.internal.matcher.DrawableMatcher.Companion.withDrawable
import com.schibsted.spain.barista.internal.matcher.DrawableMatcher.Companion.withoutDrawable

fun assertHasDrawable(@IdRes imageViewId: Int, @DrawableRes drawable: Int) {
  withId(imageViewId).assertAny(withDrawable(drawable))
}

fun assertHasAnyDrawable(@IdRes imageViewId: Int) {
  withId(imageViewId).assertAny(withAnyDrawable())
}

[...]

PR #267, opened 2018-11-15
Use top-level functions for assertions.

PaulWoitaschek:
See #260

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaAssertions.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaAssertions.kt
@@ -1,3 +1,5 @@
+@file:JvmName("BaristaAssertions")
+
 package com.schibsted.spain.barista.assertion
 
 import android.support.test.espresso.Espresso.onView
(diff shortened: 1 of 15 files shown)


Sloy:
Hi @PaulWoitaschek. Sorry for the late response 🙏 

Thanks for the PR! I just tried it locally in our project and I found a compatibility issue. Turns out that `@file:JvmName` annotation works for Java interoperability. But in Kotlin it changes the import path, from `import com.schibsted.spain.barista.assertion.BaristaVisibilityAssertions.assertDisplayed` to `import com.schibsted.spain.barista.assertion.assertDisplayed`.

It's not a bad thing by itself, but if we do it we would break the public API, which we want to avoid as much as possible.

Can you think of an alternative solution for this issue? Otherwise I think we're going to have to drop this :(

Sloy:
Also, you're right about the inconsistent formatting, my bad!
I'm going to work on https://github.com/SchibstedSpain/Barista/issues/154 to fix it

Sloy:
Since there are no new ideas, I'm closing this for now. Maybe we can apply this change in a future api-breaking version like 3.0, but not right now. 
Thanks a lot for your work @PaulWoitaschek, and sorry we didn't spot the issue before you wrote the code :(

PR #269, opened 2018-12-10
Fix assertListItemCount not throwing error when failed bug

vykintas:
Fixes bug that caused all calls to `assertListItemCount` to succeed.

Added test for this use case.

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaListAssertions.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaListAssertions.kt
@@ -12,6 +12,7 @@ import com.schibsted.spain.barista.interaction.BaristaListInteractions.findListV
 import com.schibsted.spain.barista.interaction.BaristaListInteractions.findRecyclerMatcher
 import com.schibsted.spain.barista.interaction.BaristaListInteractions.scrollListToPosition
 import com.schibsted.spain.barista.internal.failurehandler.SpyFailureHandler
+import com.schibsted.spain.barista.internal.failurehandler.withFailureHandler
 import com.schibsted.spain.barista.internal.matcher.ListViewItemCountAssertion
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/schibsted/spain/barista/assertion/BaristaListAssertions.kt
```diff
@@ -28,16 +29,18 @@ object BaristaListAssertions {
         val recyclerMatcher = findRecyclerMatcher(listId)
         val listViewMatcher = findListViewMatcher(listId)
 
-        try {
-            Espresso.onView(recyclerMatcher).check(RecyclerViewItemCountAssertion(expectedItemCount))
-        } catch (noRecyclerMatching: NoMatchingViewException) {
+        withFailureHandler(spyFailureHandler) {
             try {
-                Espresso.onView(listViewMatcher).check(ListViewItemCountAssertion(expectedItemCount))
```
Sloy: Instead of wrapping the whole method with `withFailureHandler` you can instead do `onView(...).withFailureHandler(spyFailureHandler).check(...)` in the two Espresso invocations.

Both methods work, but this one is native to Espresso, whereas `withFailureHandler` is a hack we did for some cases where Espresso doesn't support a custom failure handler. Better to avoid hacks if we can ;)

vykintas: :+1: I will update my PR with your suggestion.

vykintas:
I have updated this PR per your suggestion and this is ready to be merged.

Sloy:
Awesome! Thanks a lot for this :)

loevborg:
@Sloy when do you expect this fix to go live in a new version? 😃

Sloy:
Good point @loevborg. I think now it's a good time to do it 😛 Give me 10 minutes.

PR #285, opened 2019-03-13
Add RatingBar barista methods

alorma:
`RatingBar` is a subcomponent of `AbsSeekBar` but it works with its own methods (instead of `progress` it uses `rating`)


sample/src/androidTest/java/com/schibsted/spain/barista/sample/SeekBarTest.java
```diff
@@ -22,20 +25,39 @@
...
+
+  @Test
+  public void checkRatingBarSetProgressToMax() {
+    setRatingToMax(R.id.rating_bar);
+    assertDisplayed("7.0");
+  }
+
+  @Test
+  public void checkRatingBarSetProgressToMin() {
+    setRatingTo(R.id.rating_bar, 3);
```
alorma: This is tricky, i know

But RatingBar (and ProgressBar)  doesn't change if current value is the same as new value, and it doesn't invoke the listener, so view will not change...

PR #288, opened 2019-03-20
Add assertListNotEmpty

adrielcafe:
Because this 
```
assertListNotEmpty(R.id.my_list)
```

Is better than 
```kotlin
assertAny<RecyclerView>(R.id.my_list, "list is not empty") {
    val itemCount = it.adapter?.itemCount ?: 0
    itemCount > 0
}
```


adrielcafe:
Hey @alorma,

`assertListNotEmpty` makes total sense, my last commit renames it. 

Thanks for your suggestion!

alorma:
Thanks for your contribution

Issue #286, opened 2019-03-20
Missing assertNotDisplayed with view id and string resource


EsteveAguilera:
For the displayed assertion we do have the 
`assertDisplayed(@IdRes viewId: Int, @StringRes stringId: Int)` 

But it is missed on the not displayed assertion and currently it only accepts text
`assertNotDisplayed(@IdRes viewId: Int, text: String)`

EsteveAguilera:
This issue was fixed on #289 🎉

PR #289, opened 2019-03-25
Add assertNotDisplayed assertion with view id and string resource

EsteveAguilera:
As requested on #286 I've added the assertNotDisplayed assertion using a view Id and a String resource.


alorma:
Thanks @EsteveAguilera !

BTW, can you please update the README.md?

PR #297, opened 2019-05-14
Enable custom actions to list item child on BaristaListInteractions

gdomingues:
### What I did

Add new feature that allows us to perform a custom `ViewAction` on a child view of a list item, with a similar idea and API to the feature introduced by https://github.com/SchibstedSpain/Barista/pull/280.

I have added an example to README.md:

```java
performCustomActionOnListItemChild(R.id.list, 5, R.id.edittext, replaceText("Yet another great text"));
```

The method `BaristaListInteractions.clickListItemChild` was "rerouted" to use the newly introduced method.

`ListsChildClickTest` class was renamed to `ListsChildTest`, and tests for the new feature were added to that file.


Sloy:
Hi @gdomingues, thanks a lot for the contribution! :)
Sorry for not getting back to you sooner. It sounds like a very useful feature. Let me take a closer look to make sure it doesn't break anything weird and I'll be back.

library/src/main/java/com/schibsted/spain/barista/internal/viewaction/ChildActions.kt
```diff
@@ -0,0 +1,56 @@
...
+import android.view.View
+import com.schibsted.spain.barista.internal.failurehandler.description
+import com.schibsted.spain.barista.internal.util.ViewTreeAnalyzer
+import org.hamcrest.Matcher
+import org.hamcrest.Matchers.allOf
+
+object ChildActions {
+
+  @JvmStatic
+  fun performOnChildWithId(@IdRes childId: Int, viewAction: ViewAction): ViewAction {
```
Sloy: The `perform` word here could cause some confusion. Because this function doesn't really perform anything, it just returns the ViewAction that will be passed to Espresso's `perform()` method.

What do you think about naming it simply `onChildWithId(...)`? That way the common usage would look like `onView(...).perform(onChildWithId(R.id.child, click())`

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaListInteractions.kt
```diff
@@ -52,9 +53,15 @@ object BaristaListInteractions {
   @JvmStatic
   @JvmOverloads
   fun clickListItemChild(@IdRes id: Int? = null, position: Int, @IdRes childId: Int) {
+    performCustomActionOnListItemChild(id, position, childId, click())
+  }
+
+  @JvmStatic
+  @JvmOverloads
+  fun performCustomActionOnListItemChild(@IdRes id: Int? = null, position: Int, @IdRes childId: Int, viewAction: ViewAction) {
```
Sloy: Nice generalization! 
The name seems verbose though. I suggest a shorter version: 
```
doOnListItemChild(position = 1, childId = R.id.child, action = longClick()
```

library/src/main/java/com/schibsted/spain/barista/internal/viewaction/ClickChildAction.java
```diff
@@ -1,80 +1,20 @@
...
-import static android.support.test.espresso.matcher.ViewMatchers.hasDescendant;
-import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
-import static android.support.test.espresso.matcher.ViewMatchers.withId;
+import static android.support.test.espresso.action.ViewActions.click;
 import static android.support.test.espresso.matcher.ViewMatchers.withText;
-import static org.hamcrest.Matchers.allOf;
+import static com.schibsted.spain.barista.internal.viewaction.ChildActions.performOnChildWithId;
+import static com.schibsted.spain.barista.internal.viewaction.ChildActions.performOnChildWithMatcher;
 
 public class ClickChildAction {
```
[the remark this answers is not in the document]
gdomingues: Yeah, I saw it's no longer used, but I was afraid to remove it since it's public and as such someone could be already using it. Though if you're fine with the removal, I'll do it.

library/src/main/java/com/schibsted/spain/barista/internal/viewaction/ChildActions.kt
gdomingues: Makes sense, I'll change it 👍

library/src/main/java/com/schibsted/spain/barista/internal/viewaction/ClickChildAction.java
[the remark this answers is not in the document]
Sloy: Yes, we don't promise any api retrocompatibility for code inside the `internal` package.

gdomingues:
I tried updating my fork with the stuff that was merged to master earlier, then something apparently went wrong and now this PR is including those commits again and everything became a mess. Therefore I'll close this pull request.

PR #300, opened 2019-05-17
Update Gradle and tools

Sloy:
I upgraded the Gradle Wrapper version to the latest one. At the same time I had to update to newer versions of the Android Gradle Plugin and the Kotlin Plugin.

This fixes the error when running the project in computers with a recent version of the JDK.
`Could not determine java version from '12.0.1'.`


CristianGM:
Cheater, we approved a PR and you changed it later 😒

Sloy:
I asked on slack. That counts 😂

PR #307, opened 2019-06-06
Remove migration guide

EsteveAguilera:
Remove the outdated migration guide


alorma:
Shouldn't we add a migrate to v3?

EsteveAguilera:
There has been a long time since version 1, so I think we can assume that most projects that were using version 1 did already update to version 2...

alorma:
Makes sense...

Merge it, we should write a migration from 2 to 3. On another task

Sloy:
There isn't much to migrate for v3. Public API contracts and behaviour are maintained :shrug:

Issue #304, opened 2019-05-30
Feature request: clickOn image should be able to take a string to match contentDescription


kenyee:
**Description what you'd like to happen:**
Sometimes you want to be able to click on an image or layout w/ a contentDescription set to text because if your app was properly set up for accessibility, it would have these fields set when an image is set dynamically on it (e.g. in a Recyclerview or a dynamically created list of items whose order might change at runtime so you can't use the item index easily).
Currently, Barista doesn't have any methods to do this...I'd suggest maybe something like this for an API:

clickOnDescription(R.string.something)
clickOnDescription(textstring)

Yes, I know...do a PR :-)
Just want to finalize a possible API here before doing that...

**Example espresso code if available**
```
            onView(withContentDescription(DateCellViewHolder.getContentDescription(getApplication(), today)))
                    .check(matches(isEnabled()))
```

alorma:
I see two options here.

One is to create something like `clickByContentDescription()` that is not much `barista magic` really...

and another option is to just improve the `clickOn(R.string)``clickOn("...")` methods, to also check content description

What do you think @Sloy ?

PR #308, opened 2019-06-17
Assert content description

alorma:
This PR tries to fix: https://github.com/SchibstedSpain/Barista/issues/304

Add content description assertion, working on images, texts... any view :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,12 @@ assertHasBackground(R.id.view, R.drawable.ic_barista);
 assertHasNoBackground(R.id.view);
 ```
 
+#### Does this View have content description?
+```java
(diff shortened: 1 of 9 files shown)


alorma:
@EsteveAguilera ?

EsteveAguilera:
What about adding tests to check that it fails when the view has no content description?

EsteveAguilera:
Oh, and I've removed the content description on an ImageView to check that the tests are failing, and I've got a confusing error message:

`didn't match condition (with no content description)`

Shouldn't it be `didn't match condition (with content description)`?

alorma:
Updated!

EsteveAguilera:
👏 @alorma, now it's a 👍for me!

Issue #311, opened 2019-07-17
PermissionGranter does not work for Q devices


MatthewTighe:
**Library Version:**

3.1.0
  
**Describe the Bug:**

`PermissionGranter#allowPermissionsIfNeeded` seems to no longer work on Android Q. I _believe_ this is because the new `PermissionController` app is being used to handle permissions on Q. I can't find the source for it, so it may be impossible to refer to the "Allow" button by resource instead of by text.

I came to this conclusion by changing the code in `PermissionGranter` to the following:
```
fun allowPermissionsIfNeeded(permissionNeeded: String) {
    try {
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && !hasNeededPermission(InstrumentationRegistry.getInstrumentation().targetContext,
                        permissionNeeded)) {
            Thread.sleep(3000)
            val device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
            val allowPermissions = device.findObject(UiSelector()
                    .clickable(true)
                    .checkable(false)
                    .text("Allow"))
            val inspectPackage = allowPermissions.packageName // Inspect this in debug mode
            if (allowPermissions.exists()) {
                allowPermissions.click()
            }
        }
    } catch (e: UiObjectNotFoundException) {
    }
}
```

Inspecting the package name of the returning `UiObject` reveals that the package name is now 
`com.google.android.permissioncontroller`. 

<img width="885" alt="image" src="https://user-images.githubusercontent.com/10427470/61419617-deacf500-a8b3-11e9-9b79-47591fb4e7ac.png">

Changing the selector to use `text("Allow")` as I have in my example does fix the problem, but my assumption is that the `resourceId` is used to allow for selection in non-English locales. 

I'm happy to submit a PR for changing from `resourceId` to `text`, and I can even do it conditionally on whether the build version is Q or above if a hotfix is desirable. Otherwise the rabbit hole might be a bit too deep for me.

**Steps to reproduce the bug:**
1.  Create test to grant permissions.
2. Run test on device running Android Q.
3. Observe failure.

**Expected Behavior:**

For the permission granter to be able to grant permissions on devices targeting Q.

alorma:
hey!

Great that you can submit a PR.

Just be sure it works on other locales than english

MatthewTighe:
My solution would not work for other locales, so I will likely not be submitting a PR.

Sloy:
[...]

PR #312, opened 2019-07-19
Fix permission granter on Android Q

alorma:
Android Q has a new permission dialog package, so we needed to find which one was in order to grant permission

Important commit: 2403c8e73319f2e64d92ee5bc3c5e05d4317fbcd

Fixes https://github.com/SchibstedSpain/Barista/issues/311


alorma:
@MatthewTighe take a look!

MatthewTighe:
@alorma Brilliant! Thanks for the quick turnaround time. If you don't mind me asking, how were you able to find the new ID? I tried searching for it but was unsuccessful.

alorma:
I've used UiAutomatorView

https://developer.android.com/training/testing/ui-automator

Pretty old tool, included in tools, that it's still useful

PR #315, opened 2019-08-07
Add test for vector image assertHasDrawable

alorma:
Add missing test for `assertHasDrawable` and `assertHasAnyDrawable` with vector image check


sample/src/main/java/com/schibsted/spain/barista/sample/widget/SampleImageView.kt
```diff
@@ -0,0 +1,13 @@
+package com.schibsted.spain.barista.sample.widget
+
+import android.content.Context
+import android.util.AttributeSet
+import androidx.appcompat.widget.AppCompatImageView
+
+class SampleImageView @JvmOverloads constructor(
```
rocboronat: As this class is only intended to exist because we need it to run tests on it I would suggest adding a comment to explain it. Right now it seems a good candidate to be deleted.

rocboronat:
Oh @alorma, do you mind adding the tests that check that the assert would fail if the assertion fails?

rocboronat:
I mean doing the same that I did here: https://github.com/AdevintaSpain/Barista/pull/321

rocboronat:
@alorma Remember this PR, please! :happy

sample/src/main/java/com/schibsted/spain/barista/sample/widget/SampleImageView.kt
alorma: I've deleted it , as in fact it wasn't doing much, and will be on the library code... not needed

alorma:
Done!

rocboronat:
Thanks @alorma ! Merge at will :)

Issue #294, opened 2019-04-18
assertDisplayed with resource and string not check visibility


josinaldobarbosa:
*Please consider making a Pull Request if you are capable of doing so.*

**Library Version:**

2.10.0
  
**Describe the Bug:**

assertDisplayed(@IdRes viewId: Int, text: String) not verify if the view is visible

josinaldobarbosa:
I do not know if it is the same case, but there are others with the same behavior, for example: assertHasBackground

rocboronat:
Hello @josinaldobarbosa! I don't understand the issue. What is what you expect that is not happening? Thank you! :·)

josinaldobarbosa:
@rocboronat my bad, if the view of the parameter has visibility gone/invisibility, the test should fail

josinaldobarbosa:
@Sloy Has this been fixed?

Sloy:
Woops sorry @josinaldobarbosa, I misread you.
Would you be able to submit a test case reproducing the issue? So we're sure we understand what's the error you're describing and fix it.

josinaldobarbosa:
Sorry for the delay, but to test it is simply to put the view with invisible that the test will pass.

```
<TextView
        android:text="Hello World!"
        android:id="@+id/my_textview"
        android:visibility="gone"/>
```

`assertDisplayed(R.id.my_textview, "Hello World!")`

nickmi11er:
@Sloy @rocboronat Is there any plans to fix this issue?

rocboronat:
@nickmi11er I've just seen your PR! Thanks for it! Let's move on there 😄

Sloy:
Fixed by https://github.com/AdevintaSpain/Barista/pull/316

Thanks a lot for the report! And sorry for not responding sooner 🙏

PR #316, opened 2019-08-18
Fixed visibility assertions

nickmi11er:
This PR relates to https://github.com/SchibstedSpain/Barista/issues/294

Issue: 
assertDisplayed(@IdRes viewId: Int, text: String)
assertDisplayed(@IdRes viewId: Int, @StringRes stringId: Int)
assertNotDisplayed(@IdRes viewId: Int, text: String)
assertNotDisplayed(@IdRes viewId: Int, @StringRes stringId: Int)

check only if there is a view in the hierarchy with provided text. But it does not check that view is really (not)displayed for user


rocboronat:
I'm afraid this PR would change how Barista behaves right now, so it could break everyone's tests. Chaos! Destruction! 💣🔥💀 

What would you think about creating a new method, let's say `assertVisible` that does this new check?

@Sloy @alorma any thoughts there? Please share! 👯‍♂

glureau-betclic:
The method is in Barista**Visibility**Assertions, but doesn't handle visibility.

Also **Displayed** is pretty clear on what is expected, no one would think that a view displayed is not visible, that's weird.

The only reason to have another method name could be for retrocompatibility like @rocboronat suggested, and in this case we should **deprecate** the **assertDisplayed** and push developers to use assertVisible instead. (Even add a **assertExists** if it's needed for some people.)

glureau-betclic:
@nickmi11er maybe add tests to confirm the broken behaviour.

rocboronat:
Yes, I totally agree with you. The problem here is just a naming issue. Native Espresso names the current behavior "display". I mean, if we understand Barista as a list of shortcut recipes, `displayed` was a good name. That's why I would suggest creating a new method to avoid breaking everything.

But honestly, I know this is a mess. I'm wondering what's better for the current Barista users and I think that creating a new method to know if a `View` is actually "displayed" is the way to be. But, as _visible_ in Android means that it is _displayed_ in Espresso, it seem that the best way to handle it right now is to rename the current behavior to talk about visibility and accept the PR using the old displayed names.

Food for thought! And for the rest of Barista contributors... Guys, naming issue here! 📛

Sloy:
After writing a long paragraph about how I don't understand the issue... I understood it 😂 

Yes, you're right. The `assertDisplayed(text)` is not having the same behaviour as `assertDisplayed(id)`. It feels like a bug to me. But @rocboronat is right, changing it might break existing tests out there. 

I'm more declined towards "let's fix it and warn about it in the changelog".

josinaldobarbosa:
I always thought it worked like that, I don't see why it would "break" something

Sloy:
Hey @nickmi11er! I'm so sorry for not checking this sooner, it's been a long time. Since there has been no further discussion, I think the best choice is to merge your PR fixing the issues.

However, I'm missing some tests to make sure this doesn't break again (it should have been tested already but it wasn't). Let me add them myself and then I'll merge the PR :)

Sloy (approved):

PR #322, opened 2019-09-25
Refactor permissiongranter

inhaledesign:
I've refactored PermissionGranter to be more robust for testing.  

Summary: 
* Add instrumented tests that identify how Permissions work on Android.
* Increase legibility of code
* Increase robustness of UIAutomator code.  
  * Searching for button will timeout instead of pausing.
  * Will wait to make sure button is gone after clicking.
  * Use a regular expression to find Resource ID, which (presumably) will work if the package name changes in Android (again).

Some major issues I found when testing the Permission Granter:
1. It assumed that if a permission *has already been granted,* then it does not need to click the Grant Permission button.  It turns out this is false. Requesting a permission that has already been granted will still result in a permission dialog (tested on Android M, Pixel 3)
2. Assumed that if a permission has *not* been granted, then requesting it will show the permissions dialog.  This is false.  If a permission has not been added to the Manifest, requesting it does not show a dialog.

I've rewritten the code for the use case where the target-developer wants to test that their code correctly handles permissions. This required removing a lot to avoid false positives.  The result is more in line with a view action.

However, this has made the code no longer really fit the grantPermissionsIfNecessary() semantics.  Possibly instead of removing code, this function could also call activity.requestPermissions() so it full-on does everything for the user.  But I'm having a hard time thinking of a use-case where GrantPermissionRule would not work better.  (I also think this would not be backward compatible, since grantPermissionsIfNecessary() does not take an activity as a parameter).

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -10,6 +10,12 @@ android {
     minSdkVersion minSdkVersionDeclared
     targetSdkVersion compileSdkVersionDeclared
     testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+    testInstrumentationRunnerArguments clearPackageData: 'true'
+  }
+
+  testOptions {
+    animationsDisabled true
+    execution 'ANDROIDX_TEST_ORCHESTRATOR'
   }
 
   lintOptions {
@@ -34,6 +40,11 @@ dependencies {
   testImplementation 'pl.pragmatists:JUnitParams:1.1.0'
   testImplementation 'org.assertj:assertj-core:1.7.0'
   testImplementation 'org.mockito:mockito-core:2.13.0'
+
+  androidTestImplementation 'androidx.test:core:1.1.0'
+  androidTestImplementation 'androidx.test:runner:1.1.1'
+  androidTestImplementation 'androidx.test:rules:1.1.1'
+
 }
 
 publish {
--- a/library/src/androidTest/AndroidManifest.xml
+++ b/library/src/androidTest/AndroidManifest.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.schibsted.spain.barista">
+
+  <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+  <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+  <uses-permission android:name="android.permission.READ_CONTACTS" />
+  <uses-permission android:name="android.permission.CAMERA" />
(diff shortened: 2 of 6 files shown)


Sloy:
Hi! Sorry for the delay, and thanks for the PR :) 

I'm still reviewing the code, but in the meantime I have a question about your first point: _"Requesting a permission that has already been granted will still result in a permission dialog"_.

How did you reproduce this scenario? This is not the behaviour of Android runtime permissions. Invoking the method `Activity.requestPermission()`. When the permission is granted will **not** show the dialog again. I tested this by executing the `RuntimePermissionActivity` and clicking the "Request permission" multiple times. Here's the gif:



About point 2, good catch. We never found ourselves testing a non declared permission, but I guess it could happen. There's probably a way to check if the app has a permission declared in the manifest, we could do that and throw a descriptive error message warning the developer.

inhaledesign:
> Hi! Sorry for the delay, and thanks for the PR :)

No worries, and you're welcome.

> I'm still reviewing the code, but in the meantime I have a question about your first point: _"Requesting a permission that has already been granted will still result in a permission dialog"_.
> 
> How did you reproduce this scenario? 

[There's a test.](https://github.com/inhaledesign/Barista/blob/787775a5774c8df219fbe854b259683be4312770/library/src/androidTest/java/com/schibsted/spain/barista/internal/util/permissiongranter/PermissionGranterTest.kt#L36-L47)  I ran it on an emulated Pixel 3, Android M.  Didn't get a chance to test it on multiple devices.

> This is not the behaviour of Android runtime permissions. Invoking the method `Activity.requestPermission()`. When the permission is granted will **not** show the dialog again. I tested this by executing the `RuntimePermissionActivity` and clicking the "Request permission" multiple times. Here's the gif:
> 
> 

I wasn't expecting this behavior either, I just ran into it when I was writing the tests.  I wish I had more time to dig into this right now, but alas I won't be able to for a while.

It may be a bug or issue with [GrantPermissionRule](https://developer.android.com/reference/android/support/test/rule/GrantPermissionRule).  This seems unlikely though, since the test confirms with the activity that it has permission, before initiating a request.

I suspect it may be the _way_ in which GrantPermissionRule gives permission.  [Some permissions are still granted on installation](https://developer.android.com/training/permissions/requesting), and don't require runtime checks.  That's a pretty salient difference between your GIF and my test, is you're granting at runtime.

Or hey, maybe I made a mistake somewhere? :-)

Sloy:
I'm closing this PR because the proposed behaviour doesn't align with what we currently want PermissionGranter to do. 

Feel free to keep pushing the discussion if you disagree. Also, check out the `PermissionGranterTest` class. We improved it in https://github.com/AdevintaSpain/Barista/pull/360 and maybe it's now easier to reproduce any issues you have.

Thanks a lot @inhaledesign for your time!

PR #323, opened 2019-10-16
Add scrollTo signatures to support scrolling on custom view matcher.

glureau-betclic:
This is quite convenient when we have UI structure a bit more complex than a unique ID/string.

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/interaction/BaristaScrollInteractions.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/interaction/BaristaScrollInteractions.kt
@@ -1,9 +1,9 @@
 package com.schibsted.spain.barista.interaction
 
+import android.view.View
 import androidx.test.espresso.Espresso.onView
 import androidx.test.espresso.PerformException
 import androidx.test.espresso.matcher.ViewMatchers.withText
-import android.view.View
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/schibsted/spain/barista/interaction/BaristaScrollInteractions.kt
```diff
@@ -16,43 +16,53 @@ import org.hamcrest.Matcher
...
+        scrollWithMultipleAttempts(id.resourceMatcher(), failAtEnd = true)
+    }
+
+    @JvmStatic
+    fun scrollTo(text: String) {
+        scrollWithMultipleAttempts(withText(text), failAtEnd = true)
+    }
+
+    @JvmStatic
+    fun safelyScrollTo(matcher: Matcher<View>) {
```
rocboronat: @AdevintaSpain/barista-maintainers we should explain this method in the README or at least write a Javadoc for it...

rocboronat:
@alorma @Sloy @EsteveAguilera any feedback? IMHO we could merge it now, but I don't want to be the only eyes there. Looking for a naughty guy to dance with! 👯‍♂

alorma:
Go for it @rocboronat!

EsteveAguilera:
LGTM!

PR #325, opened 2019-10-29
Better explain how to mock the camera

rocboronat:
Right now, we're only showing the method that mocks the camera. But we don't share that the user also has to use a specific rule. This PR fixes it.

BTW, totally open to comments and suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,10 +353,12 @@ assertAny<RadioGroup>(R.id.radioGroup, "selected option is the second one") {
 }
 ```
 
-## Barista’s Intents API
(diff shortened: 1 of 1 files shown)


rocboronat:
Wep, I forgot how to mock the camera. Actually, using the Rule is not the way to do it. Give me some time to update the PR :·)

rocboronat:
And that's it! Please @alorma, check it again. Thanks! 🍺

rocboronat:
@Sloy don't be shy 😚

PR #334, opened 2019-12-02
Entry point on github actions

alorma:
From https://github.com/AdevintaSpain/Barista/issues/333

Add Github actions CI to run instrumentation tests on  every push on master and PR

### DIFF
--- a/sample/src/androidTest/java/com/schibsted/spain/barista/sample/MenuClickTest.java
+++ b/sample/src/androidTest/java/com/schibsted/spain/barista/sample/MenuClickTest.java
@@ -3,6 +3,7 @@ package com.schibsted.spain.barista.sample;
 import androidx.test.rule.ActivityTestRule;
 import androidx.test.runner.AndroidJUnit4;
 import com.schibsted.spain.barista.internal.failurehandler.BaristaException;
+import com.schibsted.spain.barista.rule.flaky.AllowFlaky;
 import com.schibsted.spain.barista.sample.util.FailureHandlerValidatorRule;
 import org.junit.Rule;
 import org.junit.Test;
@@ -52,12 +53,14 @@ public class MenuClickTest {
   }
 
   @Test
+  @AllowFlaky(attempts = 10)
   public void overflowMenuClick_byTitle() {
     clickMenu("Menu 3");
     assertDisplayed("Third menu option");
   }
 
   @Test
+  @AllowFlaky(attempts = 10)
   public void openOverflowMenu_withoutClickingAnyOption() {
     openMenu();
     assertDisplayed("Menu 3");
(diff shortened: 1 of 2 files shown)


sample/src/androidTest/java/com/schibsted/spain/barista/sample/MenuClickTest.java
```diff
@@ -46,18 +47,21 @@ public void iconMenuClick_byTitle() {
   }
 
   @Test
+  @Ignore
```
alorma: Those are flaky tests on CI, we must try to check why...

rocboronat: I think they pass in the local machine...

alorma: Yes, that's why I marked them as flaky, maybe we can merge this PR and try to fix the tests?

rocboronat: You marked them as `@Ignore`, not flaky.

Perwaps we could add a message that explains that it doesn't pass on CI but it does in the local machine? Just to help to ourselves of the future.

alorma: Let's see if `@FlakyTest` works

Sloy: Why do they fail in CI? Do we have any idea?
Maybe we could try running it locally with the exact same emulator as CI, sometimes that gives it away 🤷‍♂

alorma:
I've updated the emulator plugin version to 2.

It allows to create a matrix of emulators, cool!

There are still failing tests...

I think...

Could we merge this, and let's fix tests later? This way at least we can track failing tests, as it happened on #350 , we didn't know that some tests were failing

rocboronat:
But we're going to have tests that pass on local but not in the CI... so we'll ignore them I think 😅

alorma:
Not really.

Isn't it better to have CI with failing tests that no CI with no failing tests?

Right now on travis we are not checking that...

alorma:
I reproduced the failing tests on local.

rocboronat:
`Isn't it better to have CI with failing tests that no CI with no failing tests?`

Hum, that's ok if you had green builds before, but if you never had the suite passing, I don't see the value but having red builds on production, which is eye-painful hehe... but it's just a personal decision I think. I understand the CI as a fast way to see errors before sending things to production. If it says that everything is incorrect but actually all is working as expected, I don't know...

Sloy:
I agree with @rocboronat. Having a red build all the time doesn't help. You get used to it and ignore it :(

rocboronat:
@alorma is very near to achieve having the tests passing here. Can't wait to see them working! ❤️

alorma:
Look @Sloy @rocboronat it's green!

rocboronat:
Cool! If you think it's stable enough, feel free to merge. If it doesn't behave as we expect, we always have time to disable it. Let's try it in the real world!

Thanks for working on it, Bernat! 😊 For sure a lot of people will be "influenced" by this PR 😄

PR #336, opened 2019-12-13
Fix the matching position counter for atPosition

CemGalioglu:
The counter named "matchingPosition"  did not reset to 0, after the method "atPosition" was called once. As you can find in the test cases I included, when you create a Matcher object with atPosition, and use it once,you weren't able to use it the second time since the matcher tried to find the object again but the counter started from where it remained instead of 0, resulting in a NoMatchingViewException(in this library, it corresponds to BaristaException). Also the method in ViewTreeAnalyzer fails to find all the views the matcher compares  our target against, so I wrote a new method to find the number of views that are compared with our target.


rocboronat:
Hello @CemGalioglu! Before checking the whole PR, may you indent the code with two spaces instead of two? It's how the whole project is indented and, in addition, it will help the code comparator work.

By the way, thanks for the PR! Two days ago I was experiencing a similar issue and maybe it is related with this PR 😍

CemGalioglu:
@rocboronat  Sure, I forgot to check the result of the build after creating the PR. Is it a problem for you if I add the formatting as a new commit?

rocboronat:
No worries @CemGalioglu :D

CemGalioglu:
Did you get a chance to take a look at this? @rocboronat

rocboronat:
Sorry... I don't have any free time for Barista right now :-(

Perhaps any of the Adevinta guys could give a look to this PR? @sloydev @alorma ...?

CemGalioglu:
I realize this is a strange bug that is  somewhat "hard" to detect/fix but I believe this fix is really helpful to some people who has no idea (and will probably have no idea if they are new to testing) why their test does not work  @alorma  @Sloy  @rocboronat
Necessary tests /explanation of the bug is present in this commit, I would be more than happy to help if something is unclear (or if there is a better way to solve this) since Barista is my favorite  API in Android

rocboronat:
Oh my god, I've totally forgotten to check this PR... sorry 😣

Sloy:
Hi @CemGalioglu! So sorry for forgetting about this PR! 🙏 

I've been looking into it, and I see how the error could happen. You did an awesome job finding and fixing the issue. 👏 
I do have some conflicting opinions, though.

This matcher is in the `internal` package of the library, so it's not intended to be used outside of Barista. And it's not used anymore by the library. Since it's not part of the public API, my first instinct would be to remove it. It was first introduced to implement `clickRadioButtonPosition()`, but later it was replaced by `withParentIndex()`, which doesn't suffer from the same issue.

In general, I think Matchers that hold internal mutable state are not a good thing because they can lead to unpredictable behaviours, like the one you found. For example, using them with Barista will be a problem because Barista makes some internal retries with the same matcher, and there's no general way to "reset" those matchers.
We still have one matcher with internal state, `firstViewOf`, but it's used in a specific part of Barista were it can't be a problem.

My heart tells me to accept the PR because it fixes the issue and it's covered by tests. My head tells me to just delete the `atPosition()` matcher because we don't need it. I usually go with what my head says 😅 

The mentioned matcher, `withParentIndex()`, is a much better matcher for the same purpose. My recommendation would be to use that one instead. And if you really really need to use `atPosition` for some reason, copy the code into your own project.

Thoughts? Anyone?

rocboronat:
Hi everyone!

My heart and head say to hug those tests...!

I have to say that I don't fully understand the implementation. It's not thaaat simple... so I leave @Sloy with his war between organs. He's more aware of what's happening on Barista internally so honestly, I think his word is more important than mine.

But keep the tests, please 😄

rocboronat:
By the way, in case we go removing this `atPosition`, @Sloy what do you think about just accepting this PR and then deleting `atPosition` in another PR? Add him in the committers list is the least we can do to thank @CemGalioglu for this PR... and also his patience, hehehe...!

Sloy:
If we remove the matcher, there will be nothing for the test to test 😂

rocboronat:
...my god, that's true! All this time I thought the issue was related with an assertion and not with the matcher itself. I mean, that fixing the matcher fixed an assertion.

@CemGalioglu using Barista's regular assertions give you any issue? Now that I read twice it seems like it doesn't...

CemGalioglu:
I think I have realized this error in 2018 so so even I don't remember which method gave me the trouble back then 
Also as far as I remember One of the SO posts  linked to Barista and gave this implementation as a solution to AmbigiousViewMatcherException.

Of course this is your library and the decision to remove/keep it is up to you.

In my opinion, as someone who has created automation test frameworks similar to Barista  for huge companies with lots of developers, I would say this matcher serves to a different purpose than withParentIndex  and opening it up to external use/integrating it to functions Barista provides  is a better way to go if the purpose is to make this library even easier for people who are having difficulties with espresso to use.

Position can be an optional parameter that can be used with other matchers to combat AmbiguousViewMatcherException without knowing anything about parent views / parent in position.

A really naive/simplified example would be the following scenario.
A mock server (dedicated to testing) returns multiple texts (2 of them being 500). I have 10 instances of same CustomView in my page (of course all them have the same textview id).Clicking to text with first 500 and second 500 executes different pieces of code
If I could just write atPosition(withText("500"), 1) I would immediately get what I want without having to look up to view hierarchy. If instead I would have used something like withParentIndex() I would have to count much more and this count would always change whenever a change has been made to the layout.

Or 2 custom views that act as a checkbox with the same id corresponding to Accepting Terms/Privacy Policy in Signup Page. They will always be there and their id will always remain same so we can immediately click them. But their position in their parents will change more often.

Actually mock server isn't the best example and I can come up with much better ones but it would be way too detailed for a PR :D. (elements in ViewPagers with page margins, custom view implementations etc)

Obviously it depends on your own idea of what a library should look like, I can also help if you are convinced that adding index to matchers/assertions in Barista is a good idea as it would take some time to do it. If it were up to me I would definitely open it up to the external use as it helps in various different scenarios.

About implementation, unfortunately I don't have a "better" fix  but please teach me if you know a better way. I remember I tried so many different things but I couldn't reset the state variable to 0 at the beginning of each execution.
@rocboronat @Sloy

Sloy:
Thank you so much for that opinion @CemGalioglu, you gave me a lot to think about. 

You're right, atPosition works differently from withParentIndex. Although, when we added it, the purpose was the same. I understand the use case you're proposing, and it makes total sense.

I guess the question can be summed up as: Should Barista provide only those very specific functions listed in the readme? Or should it also provide utilities for people wanting to using a lower-level, more flexible API like Espresso? 
We've already crossed that line sometimes with functions like `assertDisplayed(matcher)`,  so I'm not sure of the answer.

Sloy:
Hi! Sorry for the long delay. 
We discussed this internally, and we decided that we don't want to add complexity to internal Matchers to support use cases outside Barista. That's the reason those Matchers are inside the `internal` package.

In your case, our recommendation would be to avoid reusing Matchers in multiple assertions.

I'm sorry it didn't go through! In any case, thank you for submitting the PR and for the discussion!

PR #339, opened 2020-03-20
add assertParentDisplayed assertion

nick2525:
Add assertParentDisplayed assertion


rocboronat:
Hello @nick2525!

Can you add a test that checks that your assertion works? And another one that tests that your assertion fails when the parent is not displayed.

Sorry for the extra work, but it's the only way we can make this project reliable. All the current actions and assertions are tested with tons of love.

Thank you!

rocboronat:
Hi @nick2525! Are you still interested in merging this PR?

By the way... why don't you give an ID to the parent and that's it? 🤷‍♂️

Sloy:
Hey @nick2525, I don't think this assertion is needed. The existing `assertDisplayed()` function already takes into account parent visibility. I.e. if the child is VISIBLE but the parent is GONE, `assertDisplayed(child)` will fail.

I'm closing this PR for now, if you have a specific scenario where this doesn't work and you can reproduce it, feel free to reopen the PR.

And thanks for contributing!

PR #341, opened 2020-04-15
Feature: Add color styleable assertTextColorIs

alorma:
Add ability to get color for `assertTextColorIs` from styleable and style resource:

``` kotlin
assertTextColorIs(
   R.id.customTextView,
   R.styleable.SampleCustomView,
   R.style.SampleCustomStyle,
   R.styleable.SampleCustomView_customColor
);
```

This way, if text color is not matching the color defined in the property `customColor`of `SampleCustomStyle` it will fail

### DIFF
--- a/sample/src/main/AndroidManifest.xml
+++ b/sample/src/main/AndroidManifest.xml
@@ -73,7 +73,8 @@
     <activity
         android:name=".ColorsActivity"
         android:exported="true"
-        />    <activity
+        />
+    <activity
         android:name=".ColorsSecondActivity"
         android:exported="true"
         />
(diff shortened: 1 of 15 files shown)


sample/src/main/AndroidManifest.xml
```diff
@@ -73,6 +73,9 @@
     <activity
         android:name=".ColorsActivity"
         android:exported="true"
+        />    <activity
```
rocboronat, line 76: We could have a checkstyle tool to check these things 😅

rocboronat:
@alorma do you think we should introduce it in the readme?

alorma:
Sure @rocboronat, I will update it, as well as changes on `assertTextColorIs` from previous PRs, but on another PR

rocboronat:
Yay @alorma, it still breaks

image

alorma:
Yes, fixing

rocboronat:
Cool! Thank you so much for this PR, @alorma :·)

PR #343, opened 2020-05-05
Exclude Room metadata when clearing DB

jmarsican:
- Provide Room metadata table name and method to exclude it from DB clearing process
- Implement test case: `doesNotDeleteRoomMetadata`
- Replace deprecated method: `InstrumentationRegistry.getTargetContext()`

Room has become the most popular ORM library and it is officially supported by Google Android team. This library creates and populates automatically a table with important database metadata: identity hash of our implemented database schema. If this metadata is deleted it can cause unexpected behavior or even crashes (`IllegalStateException`).
For more info see:
[Understanding migrations with Room](https://medium.com/androiddevelopers/understanding-migrations-with-room-f01e04b07929)
[How does Room verify data integrity](https://stackoverflow.com/questions/57547129/how-does-room-verify-data-integrity-where-are-the-db-versions-hashes-stored)
Few people is aware of this, so it is good to provide built in logic to allow easy understand and usage of this rule when using Room

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/rule/cleardata/ClearDatabaseRule.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/rule/cleardata/ClearDatabaseRule.kt
@@ -24,11 +24,6 @@ class ClearDatabaseRule(private val databaseOperations: DatabaseOperations = Dat
     return this
   }
 
-  fun excludeRoomMetadataTable(): ClearDatabaseRule {
-    excludeTablesRegex = Regex(ROOM_METADATA)
-    return this
-  }
-
   override fun apply(base: Statement, description: Description): Statement {
     return object : Statement() {
       override fun evaluate() {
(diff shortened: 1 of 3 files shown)


Sloy:
Hi @jmarsican! Thank you for your first contribution :) 

When I saw your PR I was sure we were already excluding Room metadata, because we're using Room in production and we encountered the issues you mentioned. So I checked and, we're excluding it from our app! 😱 I really thought Barista was doing it by default, but nope.

So this change is more than welcome 😄 

About your proposal, I would go even further. You're exposing a public function `excludeRoomMetadataTable()` that users can invoke. Nobody should want to delete the metadata file because it leads to issues, so let's make Barista smart by default. What do you think about always skipping that table automatically instead?

jmarsican:
Hey @Sloy I'm glad this a useful contribution. Thank you for your feedback and suggestions. I've just applied your proposal and pushed new commit.

jmarsican:
@Sloy please let me know if you want me to apply something else ;)

rocboronat:
Cool! 👏 Thank you so much, @jmarsican!

Sloy:
You should have this feature now available in the 3.5.0 release 😃

PR #347, opened 2020-05-20
Change .kotlin_module generated name

crgarridos:
Hi, I confronted the following error while building with barista and an internal company's library.

> Duplicate files copied in APK META-INF/library_release.kotlin_module

And found out that barista is taking this name too:
image

I've already fixed on our side but think this could be useful for other barista users.

------

This PR has as purpose avoid taking the name of the current folder to generate the
kotlin_module generated into the .aar/META-INF.

In practice, rename "META-INF/library_release.kotlin_module" to
"META-INF/barista.kotlin_module".

This avoid Barista consumers to exclude this file if another library
generated the same one.

See: https://stackoverflow.com/q/4450960

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -15,6 +15,7 @@ android {
   lintOptions {
     disable 'InvalidPackage'
   }
+  kotlinOptions.freeCompilerArgs += ["-module-name", "barista"]
 }
 
 dependencies {


rocboronat:
Seems legit to me 😅 Thanks for the PR! Going to try it in localhost because honesty, I (and I think we) never faced the issue 👯

rocboronat:
I didn't find any reason not to go forward with it 😄

Any feedback guys? @Sloy @alorma

rocboronat:
Oh, just read you @Sloy, didn't refresh the page.

So... merge at will then 👍

Thanks @crgarridos for this lovely contribution!

PR #350, opened 2020-05-28
Remove Jetifier

BraisGabin:
This library doesn't depend on any android support library (except glide, that it's just used in the sample app and I updated it). So we can remove jetifier safely and get faster builds 🎉.

I needed to update `mockito-android` because when I removed jetifier mockito start complaining about "final interfaces". The ways of ~the lord~ mockito are inscrutable.

BTW: there are two tests failing at `master` right now and they fail in this PR too: `checkColorList_whenDisabled` and `checkColorList_whenChecked`.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,3 +1,2 @@
-android.enableJetifier=true
 android.useAndroidX=true
 org.gradle.jvmargs=-Xmx1536m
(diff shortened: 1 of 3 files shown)


alorma:
Captura de pantalla 2020-05-28 a las 12 52 11

Uhmmm, tests are all green

rocboronat:
Cool! In this case, the less, the better.

Crossed fingers that it doesn't give any issue when releasing it 🤞 If something bad happens, @BraisGabin we'll call you on midnight! 🌙

Sloy:
Thank you @BraisGabin!

PR #351, opened 2020-05-28
Fix: Color state list now checks all possible states

alorma:
Thanks to @BraisGabin on #350 that pointed out that `checkColorList_whenDisabled` and `checkColorList_whenChecked` tests were failing.

I've investigated and found that on API 22 color state was not returning the same instance, (I believe android does some caching in future versions).

So I've moved color state list check to verify all possible states of color state list instead of checking for the same instance.

Captura de pantalla 2020-05-28 a las 14 30 18

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/internal/matcher/TextColorMatcher.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/internal/matcher/TextColorMatcher.kt
@@ -45,8 +45,15 @@ class TextColorMatcher(private val expectedColor: Int) : BoundedMatcher<View, Te
 
   private fun matchesColorList(textView: TextView): Boolean {
     val currentColorList = textView.textColors
-    val expectedColorInt = ContextCompat.getColorStateList(textView.context, expectedColor)
-    return currentColorList == expectedColorInt
+    val expectedColorList = ContextCompat.getColorStateList(textView.context, expectedColor)!!
+
+    val allStates = ALL_COLOR_STATE_LIST_STATES.flatMap { state -> listOf(state, -state) }.map { state -> intArrayOf(state) }
+
+    return allStates.all { state ->
+      val currentStateColor = currentColorList.getColorForState(state, currentColorList.defaultColor)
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/schibsted/spain/barista/internal/matcher/TextColorMatcher.kt
```diff
@@ -45,8 +45,15 @@ class TextColorMatcher(private val expectedColor: Int) : BoundedMatcher<View, Te
 
   private fun matchesColorList(textView: TextView): Boolean {
     val currentColorList = textView.textColors
-    val expectedColorInt = ContextCompat.getColorStateList(textView.context, expectedColor)
-    return currentColorList == expectedColorInt
+    val expectedColorList = ContextCompat.getColorStateList(textView.context, expectedColor)!!
+
+    val allStates = ALL_COLOR_STATE_LIST_STATES.flatMap { state -> listOf(state, -state) }.map { state -> intArrayOf(state) }
```
rocboronat, line 50: What does this `-` of `listOf(state, -state)` mean? Just curiosity

alorma: It reverses the state:

state enabled: `android.R.attr.state_enabled`
state disabled: `-android.R.attr.state_enabled`

rocboronat: WOW, that's a big WTF. What's the magic behind that?

IMHO we could do something easier to catch...

alorma: What do you mean?

alorma: https://stackoverflow.com/a/5368765/1798470

BraisGabin:
Oh! And it was a real error, i wasn't just a flaky test. Nice fix :)

PR #353, opened 2020-05-28
Update disabled states for color state list

alorma:
As @rocboronat points on #351 : https://github.com/AdevintaSpain/Barista/pull/351#discussion_r431833434

It's not clear what `"-"` does, so I've extracted states to a method and commented with link to stack overflow


rocboronat:
Thanks! It's the same WTF think but at least it's explained 😄

(I wrote _al menys_ instead of at least 😂)

PR #354, opened 2020-06-03
Fix issue #283 assertDisplayedAtPosition doesn't assert displayed lis…

### LINKED ISSUE #283: assertDisplayedAtPosition doesn't assert displayed list view
Hi,
I found a problem with `BaristaListAssertions.assertDisplayedAtPosition(...)`.
When multiple list ids match (like in a ViewPager), it scrolls the displayed list but then it asserts on the first list that matches the id.

I believe the problem is in internal `matchRecyclerView` and `matchListView` functions.
https://github.com/SchibstedSpain/Barista/blob/aed99b678889a16bbfdf08657740122698691d54/library/src/main/java/com/schibsted/spain/barista/assertion/BaristaListAssertions.kt#L108
They both use simple `view.rootView.findViewById` which returns the first match.

DeanPike:
…t view

I removed the scrollListToPosition from assertDisplayedAtPosition as it is done in assertCustomAssertionAtPosition.

isShowOnScreen checks to see if the view is visible to the user.
getShownViewsById retuns a list of all views that match the viewId and are visible to the user.

Fixes #283


Sloy:
So sorry for the delay!

I just tried it and indeed your PR fixes the issue. Thanks a lot @DeanPike!

PR #356, opened 2020-06-25
PermissionGranter should click *foreground* grant button for location

growse:
On Android Q, `"com.android.permissioncontroller:id/permission_allow_button"` isn't guaranteed to exist on permission requests where the app is requesting location permissions and only needs them for the foreground activity / service. See this example:

image

Here, `com.android.permissioncontroller:id/permission_allow_button` doesn't exist.

Looks like a small enhancement is needed to detect asking for location permission and switch to matching the "foreground only" button on the presented dialog.


Sloy:
Hello @growse, so sorry for the late reply. 
I checked and you're absolutely right, the current PermissionGranter fails when requesting location permissions in Android Q. 

I was going to ask you to add a regression test to verify the misbehavior, but I realized that the `PermissionGranterTest` is far from perfect, so please let me make some quick changes, add the missing test and then I'll merge your PR with the fix.

PR #360, opened 2020-08-04
Simplify PermissionGranterTest

Sloy:
While writing tests for https://github.com/AdevintaSpain/Barista/pull/356 I realized that this test class needed some improvements.

Here's what I did:
- Added test `ignores_already_granted_permission`. This behavior already existed but wasn't covered.
- Added test `grants_location_permission`, because in Android Q the Location permission has different buttons. This test is broken in Q, and fixed by https://github.com/AdevintaSpain/Barista/pull/356
- Deleted `fails_when_using_permission` because wasn't really testing the PermissionGranter
- Replaced `RuntimePermissionActivity` with `EmptyActivity`. By using ActivityScenario instead of BaristaRule we can request permissions from the test without having an intermediate activity with a few buttons. Now we can request any permission from the test. _Almost_ looks like a unit test.
- Used different permission for each test, to avoid issues with the test execution order.
- Deleted unused code.


Sloy:
@alorma I made some additional changes to extract some verbosity out of the tests. Do you mind checking again?

PR #363, opened 2020-08-19
Downgrade emulator version for better stability

Sloy:
I've seen some really weird test failures in our project when using API 29.

I'm not sure that they are also happening here, but since tests are failing so much in the _experimental_ CI with Github Actions, it's worth trying a downgrade to API 28.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,7 @@ jobs:
     - uses: reactivecircus/android-emulator-runner@v2.11.0
       name: Run tests
       with:
-        api-level: 29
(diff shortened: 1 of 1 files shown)


Sloy:
Coincidence? I don't think so!


image

Sloy:
Coincidence? I don't think so!

Screenshot 2020-08-19 at 11 19 36

Sloy:
Screenshot 2020-08-19 at 11 50 59

tenor-122712448

Sloy:
This time `actions/cache@v1` failed, not the tests

PR #366, opened 2020-09-09
Feature: Update PermissionGranter to allow android 11 permissions

alorma:
Updated permission granter to match ANY of the allow buttons in this order:

```kotlin
    "com.android.permissioncontroller:id/permission_allow_foreground_only_button",
    "com.android.permissioncontroller:id/permission_allow_button",
    "com.android.packageinstaller:id/permission_allow_button"
```

So we make sure that one of this buttons exists on the permission dialog.

Captura de pantalla 2020-09-09 a las 12 31 02

Also, added `allowPermissionOneTime` that accepts permission just this time, so we can do tests that involves multiple permission request time

One time permissions: https://developer.android.com/guide/topics/permissions/overview#one-time

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/interaction/PermissionGranter.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/interaction/PermissionGranter.kt
@@ -1,7 +1,5 @@
 package com.schibsted.spain.barista.interaction
 
-import android.Manifest.permission.ACCESS_COARSE_LOCATION
-import android.Manifest.permission.ACCESS_FINE_LOCATION
 import android.content.Context
 import android.content.pm.PackageManager
 import android.os.Build
@@ -16,17 +14,12 @@ import com.schibsted.spain.barista.interaction.BaristaSleepInteractions.sleepThr
 object PermissionGranter {
 
   private val PERMISSIONS_DIALOG_DELAY = 3000
-  private val PERMISSIONS_DIALOG_ALLOW_ID = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
-    "com.android.permissioncontroller:id/permission_allow_button"
-  } else {
+
+  private val PERMISSION_DIALOG_ALLOW_IDS = listOf(
+    "com.android.permissioncontroller:id/permission_allow_foreground_only_button",
+    "com.android.permissioncontroller:id/permission_allow_button",
     "com.android.packageinstaller:id/permission_allow_button"
-  }
-  private val PERMISSIONS_DIALOG_ALLOW_FOREGROUND_ID = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
-    "com.android.permissioncontroller:id/permission_allow_foreground_only_button"
-  } else {
-      PERMISSIONS_DIALOG_ALLOW_ID
-  }
-  //    private static final String PERMISSIONS_DIALOG_DENY_ID = "com.android.packageinstaller:id/permission_deny_button";
+  )
 
   @JvmStatic
   fun allowPermissionsIfNeeded(permissionNeeded: String) {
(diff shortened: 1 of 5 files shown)


library/src/main/java/com/schibsted/spain/barista/interaction/PermissionGranter.kt
```diff
@@ -36,15 +35,43 @@ object PermissionGranter {
...
+          allowPermissions.click()
         }
+      }
+    } catch (e: UiObjectNotFoundException) {
+      Log.e("Barista", "There is no permissions dialog to interact with", e)
+    }
+  }
+
+  @JvmStatic
+  fun allowPermissionOneTime(permissionNeeded: String) {
```
Sloy, line 58: If we're adding this new method it should have a test to cover it at least 😅

library/src/main/java/com/schibsted/spain/barista/interaction/PermissionGranter.kt
```diff
@@ -36,15 +35,43 @@ object PermissionGranter {
...
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && !hasNeededPermission(getApplicationContext(),
+              permissionNeeded)) {
+        sleepThread(PERMISSIONS_DIALOG_DELAY.toLong())
+        val device = UiDevice.getInstance(getInstrumentation())
+
+        val regex = PERMISSION_DIALOG_ALLOW_ONE_TIME_IDS.joinToString(
+          prefix = "^(",
+          separator = "|",
+          postfix = ")$"
+        ) { it }
```
Sloy, line 69: A big part of this code is duplicated in the two methods. It could be extracted.

sample/src/androidTest/java/com/schibsted/spain/barista/sample/PermissionGranterTest.kt
```diff
@@ -49,21 +49,21 @@ class PermissionGranterTest {
...
-            PermissionGranter.allowPermissionsIfNeeded(SIMPLE_PERMISSION_2)
+            PermissionGranter.allowPermissionsIfNeeded(CAMERA_PERMISSION)
         }
     }
 
 
 }
 
 // We can't reuse permission from one test to another, because they stay granted after each test
-private const val SIMPLE_PERMISSION_1 = Manifest.permission.READ_CONTACTS
```
Sloy, line 65: Oh! I explicitly named the permissions like this for a reason 😬 
As the comment says, each test needs to use a different permission, that's why I numbered them. I thought using the real name would make it easier to accidentally reuse them.

Sloy (approved):
Nice!

sample/src/androidTest/java/com/schibsted/spain/barista/sample/PermissionGranterTest.kt
alorma: Renamed

library/src/main/java/com/schibsted/spain/barista/interaction/PermissionGranter.kt
alorma: `ignores_already_granted_permission` uses it

library/src/main/java/com/schibsted/spain/barista/interaction/PermissionGranter.kt
alorma: done

library/src/main/java/com/schibsted/spain/barista/interaction/PermissionGranter.kt
Sloy: Hmm and I guess we can't "actually" test the behaviour of that button. I mean, check that next time the permission is not granted. Because we can't reproduce that from one test. 🤔 😞 
Well, we will have to trust Android on this one 😄

Sloy (approved):
👍

PR #368, opened 2020-09-15
Add ViewPager2 swipe support

daniellAlgar:
Hello!  

This PR will add swipe support for ViewPager2. 

It will use the exact same API as the current `swipeViewPagerForward()` and 
`swipeViewPagerBack()` but will automatically use the available ViewPager/ViewPager2. 

I ended up with a few nested `try/catch` which I'm not proud of. Please give input on it if you have a better solution. But I did see that you already have this pattern in the code base (e.g. `BaristaListAssertions`) so I thought I'd give it a try. 

Best regards,
Daniell Algar


daniellAlgar:
Pinging some of the maintainers to see that you at least have noticed this PR
@Sloy @rocboronat @alorma

Sloy:
Hi @daniellAlgar! Thanks for the PR and for the ping! I saw the notification but it got lost in my inbox, sorry 😭 

The approach you mentioned is fine by me if there's no other easy option. Let me take a look at the code to make sure there's nothing too weird :)

daniellAlgar:
Thank you @Sloy . There's no rush! I just wanted to make sure it wasn't getting forgotten :)

library/src/main/java/com/schibsted/spain/barista/internal/viewaction/SwipeActions.kt
```diff
@@ -43,6 +55,12 @@ object SwipeActions {
           NormalizedLocation(from.first, from.second),
           NormalizedLocation(to.first, to.second),
           Press.FINGER))
+
+  @JvmStatic
+  fun swipeViewPager2Forward(): ViewAction = ViewPager2SwipeAction(direction = FORWARD)
```
alorma, line 60: just thinking...

Should we add viewpager2 as part of viewpager swipe? aas other methods on barista do for list and recyclers for example

alorma (approved):

library/src/main/java/com/schibsted/spain/barista/internal/viewaction/SwipeActions.kt
daniellAlgar: I agree. It's just that swiping a ViewPager2 is a bit more complicated than swiping the old ViewPager, since ViewPager2 supports both horizontal/vertical directions (plus we always have "right-to-left" layout to consider). 

So if I were to add ViewPager2 directly in `SwipeActions.swipe(from: Pair<..>, to: Pair<..>)` I would have to figure out the pairs even before getting to this call. And that would make it even worse than what I have done today I think 🤷‍♂️ 
Now it is still the case that the user only do `swipeViewPagerForward/Back()` regardless of which view pager is being used.

But maybe I misunderstood your comment or I haven't thought this through enough :)

alorma: Good. Thanks for the explanation

daniellAlgar: I'll just add this even if it might be obvious. But a "swipe forward" means different things in ViewPager2 depending on the horizontal/vertical orientation of it. That's what I was trying to say above :)

sample/src/androidTest/java/com/schibsted/spain/barista/sample/ViewPager2Test.kt
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  @Test
+  fun swipingBackVerticallyWhileAtTheFirstPageDoesNotCrash_withId() {
+    setViewPager(orientation = VERTICAL)
+    swipeViewPagerBack(R.id.viewPager2)
+    assertDisplayed("1")
+  }
+
+  /** Helper function */
```
rocboronat, line 145: I would remove the comment. But it's not an stopper to approve the PR, of course 😄

rocboronat:
It looks great to me. Thanks for the contribution, @daniellAlgar ! 👏

rocboronat:
@alorma @Sloy time to merge?

alorma:
Yes for me @rocboronat . Go for it

Sloy:
Yes!!

rocboronat:
🎉

daniellAlgar:
Thank you all :)

alorma:
Released on [3.7.0](https://github.com/AdevintaSpain/Barista/releases/tag/v3.7.0) release

PR #371, opened 2020-09-30
Assert view background with a plain color

### LINKED ISSUE #370: assertHasBackground doesn't work with colors
**Library Version:**

3.6.0 (the latest one right now)
  
**Describe the Bug:**

When checking if a `View` has a background passing a `R.color.something`, Barista fails. This is the error it gives:

```
com.schibsted.spain.barista.internal.failurehandler.BaristaException: View (with id: com.fewlaps.app.debug:id/view) didn't match condition (with background from resource id: <2131034144>[background_warning])
```

**Steps to reproduce the bug:**
1. Make an assertion like this one: `assertHasBackground(R.id.view, R.color.background_warning);`

**Expected Behavior:**

The assertion passes if the background is applied, and fails otherwise.

rocboronat:
This PR fixes #370 

As @alorma pointed in a private secret messaging app that only we have, this solution doesn't take into account colors with states. Just plain colors. Red 🧧 Green 💚 Blue 📘

What do you think about merging this? IMHO, this PR is a small step forward ⏩

*EXTRA, EXTRA! 📰* This is how Chrome renders the *blue_car* emoji:

greenblue


library/src/main/java/com/schibsted/spain/barista/internal/matcher/BackgroundMatcher.kt
```diff
@@ -44,9 +45,16 @@ class BackgroundMatcher private constructor(@DrawableRes private val expectedDra
...
     }
 
-    val viewBitmap = DrawableToBitmapConverter.getBitmap(view.background)
-    val expectedBitmap = DrawableToBitmapConverter.getBitmap(expectedDrawable)
-    return BitmapComparator.compare(viewBitmap, expectedBitmap)
+    if (expectedDrawable is ColorDrawable) {
+      val viewDrawable = view.background as ColorDrawable
+      return viewDrawable.color == expectedDrawable.color &&
+              viewDrawable.alpha == expectedDrawable.alpha &&
+              viewDrawable.opacity == expectedDrawable.opacity
```
alorma, line 52: alpha != opacity? :S

alorma (approved):

library/src/main/java/com/schibsted/spain/barista/internal/matcher/BackgroundMatcher.kt
rocboronat: 🤷 honestly, I started comparing the values that looked important. But don't know the difference between alpha and opacity.

alorma:
Thanks for the PR @rocboronat !

PR #374, opened 2020-10-06
Is "guy" the right word?

Sloy:
I'm not a native speaker of English. But I believe "guy" is a gender-biased word. Even if we're used to seeing it as a generic word.

What do you think if we replace the Barista slogan with a different word to be more inclusive? 

From `The guy who serves a great Espresso`
to `The tool that serves a great Espresso`
or some other funnier thing.

Agree? Any better idea?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Barista
-**The guy who serves a great Espresso**
+**The tool that serves a great Espresso**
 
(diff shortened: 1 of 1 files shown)


rocboronat:
Very well spotted! As our folks suggest in the Catdroid forum, here's a +1 for `the one`.

Like Neo!

Sloy:
Great!

PR #383, opened 2021-03-02
Tech: Add material button drawable matcher

etorralbo:
Update matchesSafely implementation in Drawable Matcher class adding Material Button View.


SmasSive (approved):

Sloy (approved):
Nice job!!

rocboronat:
Hey! This PR went so, so fast 😅 

I saw there's a new test that checks that the assertion gives a green when the button has the expected value, but may we add a new test that checks the assertion fails when it doesn't?

I know we're reusing some existing code, but if we felt it was worth adding an optimistic test, I think we should also add the pessimistic one.

Thanks for the contribution, BTW!

PR #384, opened 2021-03-09
Update BackgroundMatcher to compare GradientDrawables

sarn0ld:
At the moment, when you use `assertHasBackground()` with a drawable like this:
 
```
<?xml version="1.0" encoding="utf-8"?>
<shape
    xmlns:android="http://schemas.android.com/apk/res/android"
    android:shape="oval">
    <solid android:color="@color/yellow" />
</shape>
```

then an exception is thrown because the drawable height and width are zero.

Instead of using `DrawableToBitmapConverter` a `ShapeDrawable` (=`GradientDrawable`) should be checked directly like a `ColorDrawable`.


Sloy:
Thank you for the fix @sarn0ld! 
Could you add a test for this scenario to the suite? There are some existing tests for other kinds of backgrounds using `assertHasBackground` in the `AssertionsTest` file.

rocboronat:
Hello @sarn0ld !

Thanks a lot for the PR!

As Barista is becoming a project used by thousands, may you mind adding a test that reproduces the issue you describe?

Thanks a lot again! :·)

sarn0ld:
Sure! I added tests. Thanks for the hint about the `AssertionsTest` file.

alorma:
Hi @sarn0ld 

We updated the library package, so now this PR has some conflicts that needs to be fixed before merge.

We will solve the conflicts before merge.

Sloy (approved):

PR #385, opened 2021-03-09
Add assertNoError method side of assertError methods

alorma:
When playing with errors... we have a method to check for `assertError` , error is displayed...

But, we didn't had a method to check for error is not shown...

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,6 +275,9 @@ assertHint(R.id.edittext, "Hint");
 ```java
 assertError(R.id.edittext, R.string.error);
 assertError(R.id.edittext, "Error message");
+
+assertNoError(R.id.edittext, R.string.error);
+assertNoError(R.id.edittext, "Error message");
 ```
 
 #### Check TextInputLayout's assistive helper text
--- a/README.md
+++ b/README.md
@@ -273,11 +273,11 @@ assertHint(R.id.edittext, "Hint");
 
 #### Check TextInputLayout and EditText's errors
 ```java
-assertError(R.id.edittext, R.string.error);
-assertError(R.id.edittext, "Error message");
+assertErrorDisplayed(R.id.edittext, R.string.error);
+assertErrorDisplayed(R.id.edittext, "Error message");
 
-assertNoError(R.id.edittext, R.string.error);
-assertNoError(R.id.edittext, "Error message");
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -274,6 +274,9 @@ assertHint(R.id.edittext, "Hint");
 ```java
 assertError(R.id.edittext, R.string.error);
 assertError(R.id.edittext, "Error message");
+
+assertNoError(R.id.edittext, R.string.error);
```
Sloy, line 278: Does it make sense to pass an error message to the assertion?

If I want to check that there are no errors, I would use `assertNoError()`. If I want to check that the error is "Error B" I would use `assertError("Error A")`, not `assertNoError("Error B")`.

rocboronat:
Heeey! Just wondering if `assertError` was a good name.

Maybe today I would name it `assertErrorDisplayed`, so your new method could be `assertErrorNotDisplayed`.

If you agree it, name your new method `assertErrorNotDisplayed`, and we could do the rest of the change in a new PR.

Thanks for the PR, Bernat! :-)

rocboronat (dismissed):

README.md
alorma: Done, removed

alorma:
@rocboronat renamed methods!

rocboronat:
Yey! Everything is perfect now but... there's always a but... if we merge this PR, we will break the API we offer right now.

We should mark the old `assertError` methods as deprecated and forward them internally to the new ones. We can't delete them without leaving them there for some versions. Our users will hate us if we do it. And we are lovely! 😄

library/src/main/java/com/schibsted/spain/barista/assertion/BaristaErrorAssertions.kt
```diff
@@ -1,31 +1,60 @@
...
+import androidx.test.espresso.matcher.ViewMatchers
+import com.google.android.material.textfield.TextInputLayout
 import com.schibsted.spain.barista.internal.assertAny
 import org.hamcrest.Description
 import org.hamcrest.Matcher
 import org.hamcrest.TypeSafeMatcher
 
 object BaristaErrorAssertions {
 
+  @Deprecated(
```
rocboronat, line 18: 👌

Issue #387, opened 2021-03-09
All asserts using `withText()` will fail if using on TextInputLayout


alorma:
As espresso uses:

```java
class WithTextMatcher extends BoundedMatcher<View, TextView>
```

All asserts build on top of `withText()` (like assertDisplayed(), assertContains(), ...)  will fail.

Could be a good idea to update all of this asserts with a custom `withText()` that handles `TextInputLayout` ?

Sloy:
If Espresso's matched doesn't handle it... yeah, it's probably a good idea!

PR #388, opened 2021-03-11
With text barista

alorma:
From #387 

As espresso uses:

class WithTextMatcher extends BoundedMatcher<View, TextView>
All asserts build on top of withText() (like assertDisplayed(), assertContains(), ...) will fail.

I've created a `withCompatText` that checks for `TextView` or `TextInputLayout` in order to check it's text and does the same validations of `withText`

I cannot reuse what `withText()` does, because matcher `WithTextMatcher` constructor is private


library/src/main/java/com/schibsted/spain/barista/internal/matcher/TextMatcher.kt
```diff
@@ -0,0 +1,45 @@
...
+
+fun withCompatText(string: String): Matcher<View> {
+  return withCompatText(Is.`is`<String>(string))
+}
+
+fun withCompatText(stringMatcher: Matcher<String>): Matcher<View> {
+  return withCompatTextMatcher(stringMatcher)
+}
+
+internal fun withCompatTextMatcher(stringMatcher: Matcher<String>): BoundedMatcher<View, View> {
```
Sloy, line 19: This matcher could probably be rewritten to be a composition of existing Espresso matchers instead of a copy 🤔

library/src/main/java/com/schibsted/spain/barista/internal/matcher/TextMatcher.kt
```diff
@@ -0,0 +1,45 @@
...
+      } else if (textView.transformationMethod != null) {
+        val transformedText: CharSequence = textView.transformationMethod.getTransformation(textView.text, textView)
+        return stringMatcher.matches(transformedText.toString())
+      }
+
+      return false
+    }
+
+    override fun describeTo(description: Description) {
+      description.appendText("with text on TextInputLayout: ")
```
Sloy, line 41: But, if the view is a TextView or EditText, the message will be wrong, right?

sample/src/androidTest/java/com/schibsted/spain/barista/sample/TextInputLayoutWithTextTest.kt
```diff
@@ -0,0 +1,39 @@
...
+    assertDisplayed(R.id.hintanderror_inputlayout, "Test text")
+  }
+
+  @Test
+  fun assertNoText() {
+    writeTo(R.id.hintanderror_inputlayout_editText, "Test text")
+    assertNotDisplayed(R.id.hintanderror_inputlayout, "Other test text")
+  }
+
+  @Test(expected = BaristaException::class)
```
Sloy, line 28: You could test the exception more explicitly, to check that the error message is right.
Like this one from ListClickTest:
```

  @Test
  public void fail_whenNoViewFound() {
    launchTestActivity(ListsActivity.buildIntent());

    Throwable thrown = catchThrowable(() -> clickListItem(20));

    spyFailureHandlerRule.assertEspressoFailures(1);
    assertThat(thrown).isInstanceOf(BaristaException.class)
        .hasMessage("No ListView or RecyclerView found in the hierarchy");
  }
 ```

library/src/main/java/com/schibsted/spain/barista/internal/matcher/TextMatcher.kt
alorma: resolved

library/src/main/java/com/schibsted/spain/barista/internal/matcher/TextMatcher.kt
alorma: done!

sample/src/androidTest/java/com/schibsted/spain/barista/sample/TextInputLayoutWithTextTest.kt
alorma: Uhmm, it's a generic error message.. not sure if I will add a custom message here

rocboronat:
Hello! 👋 Just a kind reminder about this lovely and very useful PR 👯

Sloy (approved):

PR #392, opened 2021-04-15
Add support to vector drawable using AppCompat

laggedHero:
Hello all,

We are using Barista and recently something caught our attention: some of our tests (API 23) where failing.

Turns out it was the way drawables are handled.

This project minSDK is 19, which means (by the way of [official docs](https://developer.android.com/guide/topics/graphics/vector-drawable-resources)) that we need to use a back compat solution. Even when considering API 23, [VectorDrawableCompat](https://developer.android.com/reference/androidx/vectordrawable/graphics/drawable/VectorDrawableCompat) docs states that `For API 24 and above, this class delegates to the framework's VectorDrawable. For older API version, this class lets you create a drawable based on an XML vector graphic.` - using a compat solution would still be needed.

Let me know whatever is needed to get this in the right state.

Thank you.


alorma (approved):
Looks good!

mustafaberkaymutlu (approved):

rocboronat:
Thank you so much, @laggedHero!

Issue #395, opened 2021-05-18
writeTo(R.id.edit_text, "String with Spaces") does not respect the EditText filter, to disallow specfic characters


ozmium:
*Please consider making a Pull Request if you are capable of doing so.*

**Library Version:**

3.9.0

```javascript
dependencies {
    ...
    androidTestImplementation('com.schibsted.spain:barista:3.9.0') {
        exclude group: 'org.jetbrains.kotlin' // Only if you already use Kotlin in your project
    }
}
```

**Describe the Bug:**

Layout has:

```xml
<com.google.android.material.textfield.TextInputEditText
    android:id="@+id/edit_text"
    android:layout_width="match_parent"
    android:layout_height="wrap_content"
    android:focusableInTouchMode="true"
    android:hint="Type some text with spaces"
    android:digits="ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz1234567890"
    android:inputType="textFilter"
    android:singleLine="true"
    android:imeOptions="actionNext" />
```

Notice the `android:digits` and `android:inputType` which create a filter to allow or disallow certain characters. Spaces are disallowed. See: https://stackoverflow.com/questions/33993041/android-disable-space-only-for-edittext

`writeTo(R.id.edit_text, "String with Spaces")` adds spaces to the EditText which has disallowed spaces.

Unit test has:

```kotlin
// Barista methods:
writeTo(R.id.edit_text, "String With Spaces")
assertDisplayed(R.id.edit_text, "StringWithSpaces") // assertion fails

// Espresso methods:
onView(withId(R.id.edit_text)).perform(typeText("String With Spaces"))
onView(withId(R.id.edit_text)).check(matches(withText("StringWithSpaces"))) // assertion passes
```

**Steps to reproduce the bug:**
1. Create an `EditText` or `TextInputEditText` which uses a filter and restricted digits, like the example above. See https://stackoverflow.com/questions/33993041/android-disable-space-only-for-edittext
2. Write a Barista test to type characters into this EditText. The characters should include some of the disallowed characters, like spaces.
3. Assert with Barista that the EditText does not contain the disallowed characters

**Expected Behavior:**

Barista should not insert the disallowed characters, because a user would not be able to do this with the on-screen keyboard. The Barista assertion for the correct string should pass. Barista's `writeTo()` should behave like Espresso's `typeText()`.

**Actual Behavior:**

[...]

PR #396, opened 2021-05-19
Replace replaceText with typeText

alorma:


rocboronat:
Hey @alorma ! What do you think about #397 ? Personally I think it's better to offer both actions. Well, answer me in #395 , to avoid duplicating the discussion 😅

PR #397, opened 2021-05-19
Fix Issue #395: Add BaristaEditTextInteractions.typeTo()

ozmium:
### BaristaEditTextInteractions

* Add a new `typeTo()` Text Interaction in `BaristaEditTextInteractions`. It's similar to `writeTo()`, but it uses Espresso's `typeText()` instead of `replaceText()`.
* Add KotlinDoc/JavaDoc to the Text Interaction methods, for users to tell the difference between `writeTo()` and `typeTo()`.

--------

### activity_edittext.xml

* Add 3 new constrained `EditTexts` to the layout, for UI unit tests of `typeTo()`
* The `EditTexts` have maxLength of 10 characters, and don't allow spaces, and only allow ASCII text or digits 0-9
* More info about `EditText` filters: https://stackoverflow.com/questions/33993041/android-disable-space-only-for-edittext

--------

### EditTextTest (UI Instrumentation Tests)

* Check that `typeTo()` will constrain and limit the typed characters to ASCII and digits 0-9, as per the EditText `digits` attribute.
* Check that `typeTo()` will limit the length of the typed text to 10 characters, as per the EditText `maxLength` attribute.
* Check that `typeTo()` will disallow typing spaces into the `EditText`, if the `textFilter` does not allow it.
* Check that the auto-scroll to the `EditText` will work when `typeTo()` is called.

### DIFF
--- a/library/src/main/java/com/schibsted/spain/barista/interaction/BaristaEditTextInteractions.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/interaction/BaristaEditTextInteractions.kt
@@ -17,7 +17,7 @@ object BaristaEditTextInteractions {
   /**
    * This uses [androidx.test.espresso.action.ViewActions.replaceText],
    * to remove any existing text and insert the characters directly,
-   * disregarding any EditText properties like `maxLength`, `textFilter`, etc.
+   * ignoring any EditText properties like `maxLength`, `textFilter`, etc.
    */
   @JvmStatic
   fun writeTo(@IdRes editTextId: Int, text: String) {
--- a/library/src/main/java/com/schibsted/spain/barista/interaction/BaristaEditTextInteractions.kt
+++ b/library/src/main/java/com/schibsted/spain/barista/interaction/BaristaEditTextInteractions.kt
@@ -3,6 +3,7 @@ package com.schibsted.spain.barista.interaction
 import androidx.annotation.IdRes
 import androidx.test.espresso.action.ViewActions.clearText
 import androidx.test.espresso.action.ViewActions.replaceText
+import androidx.test.espresso.action.ViewActions.typeText
 import androidx.test.espresso.matcher.ViewMatchers.isAssignableFrom
 import androidx.test.espresso.matcher.ViewMatchers.isDescendantOfA
 import androidx.test.espresso.matcher.ViewMatchers.withId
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/schibsted/spain/barista/interaction/BaristaEditTextInteractions.kt
```diff
@@ -13,6 +14,11 @@ import org.hamcrest.Matchers.anyOf
 
 object BaristaEditTextInteractions {
 
+  /**
+   * This uses [androidx.test.espresso.action.ViewActions.replaceText],
+   * to remove any existing text and insert the characters directly,
+   * disregarding any EditText properties like `maxLength`, `textFilter`, etc.
```
rocboronat, line 20: As a non-native English speaker, I would replace `disregarding` with `ignoring`. I think it means the same while it's easier to understand. But I could also study a bit more... Any thoughts there? Do you think this suggestion is valid?

rocboronat (approved):
OMG, this is the best-written PR I've read I think 😅 Thanks a lot for the detailed explanation, @ozmium!

The only thing this PR misses is to add this new action to the readme file. Of course, we can do it later if you want to. You made the difficult part 😄

alorma (approved):
Thanks

Sloy (approved):
Thanks for the contribution @ozmium! And for such a detailed PR description, helps a lot with the review 😃

Sloy:
image

Nooo! I didn't! GitHub did it for me! 😱 
Sorry, can you approve again? @alorma @rocboronat

Sloy (dismissed):

alorma (dismissed):
¬¬ @Sloy

rocboronat (dismissed):

library/src/main/java/com/schibsted/spain/barista/interaction/BaristaEditTextInteractions.kt
ozmium: Yes I think it's good to change to the shorter word.

rocboronat:
I would say:

```
writeTo(R.id.edittext, "A great text"); // Ignores the EditText restrictions like maxLength and textFilter. It's blazing fast.
typeTo(R.id.edittext, "A great text"); // Honors the EditText restrictions like maxLength and textFilter. It slows down the test.
```

Maybe it's too long?

Sloy (approved):

rocboronat (approved):
Thanks @Sloy!

alorma (approved):

PR #398, opened 2021-05-20
Replace "we" with "Barista"

rocboronat:
It's strange to talk about us as developers instead of Barista as software. Barista will do the hard job. We are drinking margaritas in an old lost paradisiac island 🍸


alorma (approved):

Sloy (approved):
Good point!

PR #399, opened 2021-05-21
Remove Bintray's badge

rocboronat:
It gives a 404. So ugly. Do we have to replace it with anything? I think not, but asking is always safer 😄


alorma:
No until we migrate to mavenCentral....

alorma (approved):

rocboronat:
The build passed at last! 🎉 🌮

PR #405, opened 2021-05-27
Update Gradle and AGP

Sloy:
- Updated Gradle 7
- Updated Kotlin 1.4
- Removed Jacoco task because of incompatibility. And we're not using it anyway.
- Removed Novoda bintray plugin. It's not compatible with newer versions of Gradle and AGP, and we can no longer publish to JCenter. We will have to reimplement the publishing for mavenCentral.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,20 +1,18 @@
 buildscript {
-  ext.kotlin_version = '1.3.31'
   repositories {
     google()
-    jcenter()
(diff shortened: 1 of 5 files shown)


rocboronat:
What a clean! 👌

rocboronat (approved):

alorma (approved):

Issue #403, opened 2021-05-27
Move the ListView asserts in the interactions section of the readme to the assertions section


rocboronat:
![image](https://user-images.githubusercontent.com/456499/119799479-661a1500-bedc-11eb-83cc-880fd5d0971a.png)

PR #406, opened 2021-05-27
Move the List assertions from the interactions section to the assertions one

rocboronat:
Just fixing #403


rocboronat:
This is a duplicated of #407, which in addition, is better 😄

PR #417, opened 2021-05-27
Parallelise Github actions

Sloy:
We have 2 actions, one for the build and another for running instrumentation tests.
Right now, the instrumentation action waits for the build to finish first. It takes about 2 minutes. But there's no need for that, really. There's no dependency between them or any reused output. So we could save 2 minutes by starting the instrumentation build as soon as possible.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,7 +29,6 @@ jobs:
         run: ./gradlew assembleDebug
 
   test:
-    needs: build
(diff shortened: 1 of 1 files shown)


alorma (approved):
My bad 😅😅

rocboronat:
I guess that's because it was great to save some cpu cycles passing the instrumentation tests in case the unit tests don't pass. When we did it I guess we were limited to a number of daily instrumentation tests or something like that, like in the free tier of Firebase.

PR #422, opened 2021-06-08
Update build.gradle files to gradle.kts

alorma:
This migration is not mandatory.. but will make the process of migration to **mavenCentral** lots easier...


sample/proguard-rules.pro
```diff
@@ -2,7 +2,7 @@
 # By default, the flags in this file are appended to flags specified
```
Sloy, line 2: You could just remove this file 😆

Sloy (approved):

sample/proguard-rules.pro
alorma: removed

sample/proguard-rules.pro
```diff
@@ -2,7 +2,7 @@
...
 # in /Users/roc/.android-sdk/tools/proguard/proguard-android.txt
 # You can edit the include path and order by changing the proguardFiles
-# directive in build.gradle.
+# directive in build.gradle.kts.kts.
```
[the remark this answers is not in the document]
alorma: Uhmmmmm, no. But anyway, this file is not needed

rocboronat (approved):

PR #423, opened 2021-06-08
Move package to com.adevinta.android

Sloy:
In light of the mavenCentral migration, we're moving the library package to an updated `com.adevinta.android` instead of the old Schibsted package.

Consumers will need to find&replace their imports thought. It was now or never!


rocboronat (approved):
Now or never! ✊

alorma (approved):
I guess....

Issue #431, opened 2021-08-09
Update deprecated librearies


JJaviMS:
Hello,

I had compiled the project and I had a lot of deprecation warnings, I will try to fix them.

lordcodes:
Is this issue fixed, as the PR was merged?

Sloy:
Yes! Thanks again :)

Issue #430, opened 2021-08-01
Upgrade to latest Espresso version


VictorIreri:
This brilliant library is [using Espresso 3.1.1](https://github.com/AdevintaSpain/Barista/blob/master/library/build.gradle.kts) which was released in 2018. I think it's about time it was updated to a later version, ideally the latest **3.4.0** https://github.com/android/android-test/releases/tag/androidx-test-1.4.0

Sloy:
Totally right! I can't do this right now, but a Pull Request would be welcomed :)

lordcodes:
I believe this can be closed, as the PR was merged.

Sloy:
@lordcodes yep, thanks for the head's up :)

VictorIreri:
Thanks @JJaviMS!

PR #432, opened 2021-08-09
Update Espresso version

JJaviMS:
Hi,

According to #430 I have updated the Espressor dependencies version. Due to a Dex error, caused by a too big Dex file, I had to add Multidex to the sample app.

I have also started to update some other versions, like the AndroidJUnit4 test runner and the Kotlin version (#431).


sample/build.gradle.kts
```diff
@@ -12,6 +12,7 @@ android {
   defaultConfig {
     minSdk = 19
     targetSdk = 30
+    multiDexEnabled = true
```
Sloy, line 15: Why was multidex enabled? Is the sampel app too big after the dependency changes?

If that's the case, it's probably easier to just bump the minSdk version to `21` and forget about multidex.

library/build.gradle.kts
```diff
@@ -24,10 +24,12 @@ android {
 }
 
 dependencies {
-  api("androidx.test.espresso:espresso-core:3.1.1")
-  api("androidx.test.espresso:espresso-contrib:3.1.1")
+  api("androidx.test.ext:junit:1.0.0")
```
Sloy, line 27: I'm curious, why was this dependency added?

Sloy (commented):
Hello @JJaviMS!
Thanks a lot for the PR. A dependency update was long overdue :)

I just added a couple of questions to make sure.

sample/build.gradle.kts
JJaviMS: I did not know that using minSdk into 21 would fix that issue, if it does then I will bump the minSdk version and remove multidex!

library/build.gradle.kts
JJaviMS: I added that depency because I also updated the AndroidJUnit runner in androidx.test.runner.AndroidJUnit4 is deprecated and the new one is into androidx.test.ext.junit.runners.AndroidJUnit4. which depends on androidx.test.ext:junit.

sample/build.gradle.kts
Sloy: Yes please, that would be great. Min SDK 21 is good enough for the sample.

library/build.gradle.kts
Sloy: I think we can avoid this dependency altogether if we remove all those `@RunWith(AndroidJUnit4::class)` annotations from the tests. They are not really needed.
Then, this dependency won't be needed, right? Do you want to do it? I can send a commit otherwise.

JJaviMS:
Hi!

I have removed the multidex and update the min SDK to 21. I just left the Barista minSdk in 19, but maybe we could bump it into 21 too.

Also I have removed the AndroidJunit4 runner from all the test.

Finally I have fixed some assertions test, those test failed because after updating the Espresso version some assertions messages changed, so I have changed all those messages.

PR #433, opened 2021-08-19
Fix readme for new artifact name

ptornhult:
Was a bit confusing, but after checking the release note I managed to get the new dependency working without jcenter. Was just a small miss in the readme I think. Nice work!


Sloy:
Our bad! We too excited about the migration and copied the wrong name 😅 

Thanks a lot for the fix, and sorry for the confusion ❤️

Sloy (approved):

alorma (approved):

PR #443, opened 2021-09-09
Tech: Get name for artifact from each module

alorma:
When we updated our release process to work with sonatype, we had to add `libArtifact` to **gradle.properties**, but that not allowed us to deploy different modules to maven central.

This PR allows us to have different modules, so we can deploy #439 (barista-compose) someday!


rocboronat:
Cool 👍

PR #446, opened 2021-12-01
Api 12

alorma:


sample/src/main/AndroidManifest.xml
```diff
@@ -17,128 +17,128 @@
       >
     <activity
         android:name=".SomeViewsWithDifferentVisibilitiesActivity"
-        android:exported="false"
+        android:exported="true"
```
rocboronat, line 24: Hi @alorma ! Just a question. Why do we need to set this as `true`? It's something related to passing the tests?

alorma: just trying things... tests are not passing on android 12 and i'm investigating why

PR #450, opened 2021-12-16
Tech: make library compile and target Android 12 (API 31)

alorma:


IanArb:
Hey @alorma would you like me to finish this work? If I migrated my project to Android 12 today would this library prevent me from doing so?

alorma:
Sure. If you have time, go ahead with it

PR #454, opened 2022-02-25
Compile library with API 31

IanArb:
* Update CameraActivity.kt to handle camera permissions
* Update CameraTest.java to handle case for deny permission


IanArb:
Hey @alorma @Sloy I managed to get it compiled against Android 12. However, there are some tests that were failing in previous builds and they are ok running locally. It seems to be trying failing due to the material components. Would this be considered a blocker for this if these tests are not passing?
`
No view matching ((an instance of android.widget.TextView and view.getText() with or without transformation to match: is "Hi! I'm the second screen!" or (view.getParent() is assignable from class <class com.google.android.material.textfield.TextInputLayout> and (view is an instance of android.view.ViewGroup and has descendant matching an instance of android.widget.TextView and view.getText() with or without transformation to match: is "Hi! I'm the second screen!")))) was found`

sample/src/main/java/com/adevinta/android/barista/sample/CameraActivity.kt
```diff
@@ -1,46 +1,79 @@
...
+                    this,
+                    Manifest.permission.CAMERA
+                ) == PackageManager.PERMISSION_GRANTED
+            ) {
+                capturePhoto.launch(getPictureUri())
+            }
+            else {
+                // You can directly ask for the permission.
+                // The registered ActivityResultCallback gets the result of this request.
+                requestPermissionLauncher.launch(
```
rocboronat, line 63: Do we really need this change? In the previous version, if the user didn't give the permission, the camera launch call broke, and that's great to capture it from the test side. WDYT? It's just to avoid messing with something it worked. I get this new version lets you, as a user, interact with this Activity, but that's not supposed to happen. This Activity exists only to check Barista permissions handler works as expected.

sample/src/androidTest/java/com/adevinta/android/barista/sample/CameraTest.java
```diff
@@ -25,6 +31,22 @@ public void takePictureAndShowIt() {
...
+    PermissionGranter.allowPermissionOneTime(Manifest.permission.CAMERA);
+
+    // Allow time for the assertion to complete
+    sleep(500);
+
     assertHasAnyDrawable(R.id.image_view);
   }
+
+  @Test
+  public void denyPermissionForCamera() {
```
rocboronat, line 43: Having this new check is great. Thanks! 😄

sample/build.gradle.kts
```diff
@@ -7,11 +7,11 @@ plugins {
 apply(from = "../config/android-quality.gradle")
 
 android {
-  compileSdk = 30
+  compileSdk = 31
```
rocboronat, line 10: ❤️

rocboronat:
> Hey @alorma @Sloy I managed to get it compiled against Android 12. However, there are some tests that were failing in previous builds and they are ok running locally. It seems to be trying failing due to the material components. Would this be considered a blocker for this if these tests are not passing? ` No view matching ((an instance of android.widget.TextView and view.getText() with or without transformation to match: is "Hi! I'm the second screen!" or (view.getParent() is assignable from class <class com.google.android.material.textfield.TextInputLayout> and (view is an instance of android.view.ViewGroup and has descendant matching an instance of android.widget.TextView and view.getText() with or without transformation to match: is "Hi! I'm the second screen!")))) was found`

If they didn't work on previous versions, it won't be a blocker to this PR. Don't worry! 😆 But feel free to open an issue, now that you've seen which tests didn't pass. Just to know the homework we have to do 😃 

About my feedback, it's just a comment. I love having as simpler as possible Activities, so any code we could remove, I love the idea. That's why I think we could leave the CameraActivity as it was if we can, to keep this PR minimal.

sample/src/main/java/com/adevinta/android/barista/sample/CameraActivity.kt
IanArb: Ah ok. I didn't have much context of what the purpose of the CameraActivity is. However, I assumed it was broken and I guess this is intentional? If so I can revert the changes to keep the PR minimal.

rocboronat: You are totally right. We should explain it in the contributors page. I opened an issue to remember it: https://github.com/AdevintaSpain/Barista/issues/455

Sorry for the invested time we're going to waste! 😞

IanArb: No worries at all. Thanks for clarifying.

IanArb:
Hey @rocboronat I've reverted the camera activity changes and created an issue for the failing tests. https://github.com/AdevintaSpain/Barista/issues/456

rocboronat:
Great! I'll take a last look and merge it. Thannnnks a lot! 😊

IanArb:
> Hey @rocboronat I've reverted the camera activity changes and created an issue for the failing tests. #456

Hey @rocboronat any follow up on this review?

alorma:
Hi @IanArb .

We've updated the library to API 33.

PR #457, opened 2022-04-12
Tech: Add drawable matcher extensions file with an example test

etorralbo:
Added on node with drawable matcher and included an example of how to use it.


barista-compose/src/main/java/com/alorma/barista_compose/assertion/internal/matcher/DrawableMatcher.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import androidx.annotation.DrawableRes
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.semantics.SemanticsPropertyKey
+import androidx.compose.ui.semantics.SemanticsPropertyReceiver
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.test.SemanticsMatcher
+import androidx.compose.ui.test.junit4.ComposeTestRule
+
+fun Modifier.setDrawableSemantic(@DrawableRes id: Int): Modifier {
```
alorma, line 11: The idea is to set this modifier on production code?

alorma: uhmm, then the code for the modifier shouldn't go on the same module than barista, as barista is only used for tests (`andoridTestimplementation`)

alorma: maybe we need ... "barista-helpers"  or something

etorralbo: OK! I had doubts because I've had to load this module in the project to run the test. So, you propose to create a new one called barista-helpers and move there the DrawableMatcher file, don't you?

PR #461, opened 2022-10-11
Improve assertion error messages by adding mismatch descriptions

angusholder:
I added overrides for TypeSafeMatcher.describeMismatchSafely() to improve the error messages that Barista gives when the assertion fails


alorma:
Hi @angusholder 

Any example on how error messages looked before this PR and how it looks with this change?

Sloy:
Weren't there some tests for the assertion messages? I don't know what's the current status of CI

alorma:
Look much better!

Is there any way, btw, to avoid this ñart?

> Caused by: junit.framework.AssertionFailedError: 'with error: "My error message 123"' doesn't match the selected view.

alorma (approved):

angusholder:
> Look much better!
> 
> Is there any way, btw, to avoid this ñart?
> 
> > Caused by: junit.framework.AssertionFailedError: 'with error: "My error message 123"' doesn't match the selected view.

What do you want it to say instead? I think that part is created by JUnit and Hamcrest

PR #467, opened 2022-11-22
Move fake secrets to local.properties

alorma:
Remove fake signature variables from publish-root.gradle


rocboronat:
Yey @alorma, a comment from the outside 😅 Now you are the only one that knows how the publication process works. Before this PR, people had a clue about how it worked in case you are not here to launch a new version. But now there's more mystery to solve. Do you think you could explain how the publication works somewhere? I know it's really boring stuff but if you could do it, it would be super appreciated by someone eventually, haha!

rocboronat:
Thank you so much, and sorry for the extra stuff to do...!

PR #472, opened 2022-12-02
Fix code style script

Sloy:
Same as https://github.com/AdevintaSpain/TaggingViewer/pull/5


rocboronat:
Yey, I would say yes to the first comment deletion, but the second one... well, with it you know there are .xml files there 😅 

Do we really care about those comments? Or did they actually break the script?

rocboronat (approved):

alorma (approved):

PR #479, opened 2023-01-03
Extend assertNotExist with a new implementation using a viewMatcher as a param

moutyque:


rocboronat:
Hello @moutyque 👋

Thanks for the PR! It looks good to me, although I'm a bit disconnected from the project, so I would like to have another reviewer here, maybe @alorma :)

One thing, could you also add it to the readme?

Thank you so much!

rocboronat:
By the way, checking the current readme we have, if we start implementing things like this, we should also add the rest of matchers, right? I mean assertExists, assertVisible... etc...

moutyque:
Hello @rocboronat ,
Thx for your quick feedback.
About implementing other method as well, I do believe it is a good idea however I would like to do that in different PR to avoid being stuck for too long. Doing one function per PR seems to me a better way to do it because we could provide new features over the time and not having to wait a big batch to release them all.
Please give me your feedback.

moutyque:
Also I run locally the step that failed with the CI: ./gradlew :sample:connectedCheck :barista-compose:connectedCheck
and it ends properly. Could you advice how to fix the CI ?

alorma (commented):
Hi @moutyque! Sorry for the delay on my review. I was out of laptop for holidays.

Thanks for this PR. I think it could be very useful!

Btw, as I agree with @rocboronat on other methods, I will suggest to create at least the `assertExists` variant also, so we have both positive and negative versions. Later another PR for other kind of methods.

What do you think?

alorma (approved):

moutyque:
Hello @alorma , 
I fact I though assertExist already exist but no.
I agree with you, I will add the positive assertion.

alorma (approved):

PR #501, opened 2023-07-11
Bump multiple dependencies

vudzkostek:
Dependency bump per commit.

Sample tests went through, also tested in commercial app that everything still works fine.


sample/src/androidTest/java/com/adevinta/android/barista/sample/util/MaterialAttributes.java
```diff
@@ -0,0 +1,6 @@
+package com.adevinta.android.barista.sample.util;
+
+public interface MaterialAttributes {
```
vudzkostek, line 3: Because project is using Java, we could not use alias, but I also wanted to avoid inlined packages

vudzkostek: Explanation: latest gradle enabled by default `nonTransitiveRClass`

